PR #2, opened 2020-03-22
Remove dependency on ansible.netcommon collection

felixfontein:
##### SUMMARY
Right now the community.crypto collection has a dependency on [ansible.netcommon](https://github.com/ansible-collections/netcommon/) only because of [module_utils/compat/ipaddress.py](https://github.com/ansible-collections/netcommon/blob/master/plugins/module_utils/compat/ipaddress.py).

While I generally don't like duplicating code, I don't see why we should have a network device collection as a dependency only for one file which only "randomly" ended up in that collection (it's a vendored version of https://pypi.org/project/ipaddress/).

CC @ctrufan @MarkusTeufelberger @puiterwijk @Shaps @Spredzy @Xyon

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/compat/ipaddress.py

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -8,8 +8,6 @@ description: null
 license: GPL-3.0-or-later
 license_file: COPYING
 tags: null
-#dependencies:
-#  ansible.netcommon: '>=0.1.0'
 repository: https://github.com:ansible-collections/community.crypto
 documentation: https://github.com/ansible-collection-migration/community.crypto/tree/master/docs
 homepage: https://github.com:ansible-collections/community.crypto
--- a/plugins/module_utils/acme.py
+++ b/plugins/module_utils/acme.py
@@ -31,7 +31,7 @@ import traceback
 from ansible.module_utils.basic import missing_required_lib
 from ansible.module_utils._text import to_native, to_text, to_bytes
 from ansible.module_utils.urls import fetch_url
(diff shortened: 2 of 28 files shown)


felixfontein:
Tested with:
 * `ansible-test integration -v --truncate=0 --docker-no-pull --docker ubuntu1804`
 * `ansible-test units --python 3.7 --docker default --color -v`
 * `ansible-test sanity -v --truncate=0 --docker-no-pull --docker`

gundalow:
Should we add a note in the file in `ansible.netcommon` to say we've forked it? That might help protect us if someone makes an updated to the original version?

MarkusTeufelberger:
Isn't the "original version" actually https://pypi.org/project/ipaddress/ anyways?

gundalow:
> Isn't the "original version" actually https://pypi.org/project/ipaddress/ anyways?

Yup, though in ansible.netcommon would help a bit. 
Either way, if the maintainers of `community.crypto` are happy then I'm +1 to this PR

MarkusTeufelberger:
"Friendship ended with `ansible.netcommon` - Now upstream is our best friend!" :-P

In all seriousness though a note might help the `netcommon` team to track if they even need to keep these vendored files around.

felixfontein:
There are other modules in community.general which use ipaddress from ansible.netcommon.

PR #4, opened 2020-03-24
Run tests in Shippable

felixfontein:
##### SUMMARY
Mainly copied the config from community.general and trimmed it down.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
CI


felixfontein:
This probably fails until #2 is merged.

felixfontein:
The integration tests probably need ansible/ansible#67684 before they can be run. Sanity tests essentially work, except they currently also check the files in tests/utils/shippable, which make it fail... (Same problem in community.general.)

felixfontein:
I've moved the workarounds to #5 and will continue on them after merging this. Right now it works everywhere but on OpenSUSE and FreeBSD, since their package managers are in community.general and not in ansible/ansible, and it will properly be fixed in ansible/ansible#67684.

PR #7, opened 2020-03-29
Rename openssl_certificate* to x509_certificate*

felixfontein:
##### SUMMARY
The discussion in ansible/ansible#63801, whether to rename some/most/all of the openssl_* modules to x509_*, never got any traction. So now here's the most basic version of that proposal (original suggested in ansible/ansible#33715): let's rename `openssl_certificate*` to `x509_certificate*` (both the regular and the `_info` module).

What do you think?

This should have no impact on our users (assuming we adjust the routing.yml file in ansible/ansible, which I'll do once this is merged), and before a definite first release of community.crypto, changing its structure is still trivial. (Soon there will be a local routing file per collection, where such renames can be recorded. But right now it's even simpler :) )

Fixes ansible/ansible#33715, closes ansible/ansible#63801.

@ctrufan @MarkusTeufelberger @puiterwijk @Shaps @Spredzy @Xyon What do you think? Should we do this now, later, or never? :)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
openssl_certificate
openssl_certificate_info


Shaps:
I believe if we want to make this change ( and I think we should ), now is the right time!

felixfontein:
@ctrufan @MarkusTeufelberger @puiterwijk @Spredzy @Xyon any last opinions? If nobody writes anything until this weekend, I'll merge this.

MarkusTeufelberger:
My opinion hasn't changed - I'm fine with it either way and the old name will keep working (with a deprecation notice) for a while anyways.

felixfontein:
Deprecation of the old name is a good point. I'm not really sure how (well) that will work in collections. It's something I will look at once the (in-)famous ansible/ansible#67684 gets merged :)

The old (short) name will continue to work in 2.10 anyway if we update routing.yml (which I plan to do right after merging this).

felixfontein:
I've been playing around with deprecation and redirection a bit (I'm using ansible/ansible#67684), and in the end decided to add 04e46506e4de9d7d8df6b8ed9c41ced99ce28dc9 and 101fd3bd1590d707ce6ec7722ba98cdd4c97355a.

The results are as follows:

1. If the user uses `openssl_certificate` and `openssl_certificate_info` (short names), they will work as before. No warnings will be printed, and they can still be used in the future. (Ansible's routing file needs to be adjusted for this. Will create a PR once this is merged.)

2. If the user uses `community.general.openssl_certificate` or `community.general.openssl_certificate_info` (FQCNs), a warning will be printed by Ansible: `[WARNING]: [DEPRECATION WARNING] The 'community.crypto.openssl_certificate' module has been renamed to 'community.crypto.x509_certificate'`. (This is caused by the entry in the `meta/routing.yml` file in this collection.) Note that this **will only work for Ansible 2.10 or newer!** Also, note that this warning will be printed prominently at the beginning of the playbook (when using static imports I guess).

3. Especially for users of **Ansible 2.9**, another warning will be printed when using the modules for the first time in a playbook (i.e. regular behavior of Ansible warnings) when `community.general.openssl_certificate` or `community.general.openssl_certificate_info` is used:
`[DEPRECATION WARNING]: The 'community.crypto.openssl_certificate' module has been renamed to 'community.crypto.x509_certificate'. This feature will be 
removed in version 2.14. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.`

4. If the user uses `community.general.x509_certificate` or `community.general.x509_certificate_info`, everything works fine and no warning is printed.

@MarkusTeufelberger @Shaps @resmo do you think this is OK?

felixfontein:
I added a paragraph to the module documentation explaining this (d0e60c369cd48d4eff237dbb6617ed1b6d388cab).

MarkusTeufelberger:
Yeah, I think that's mostly what users will expect. The actual details of how deprecation works (date based, collection release based, ansible release based) will anyways still be decided or can be decided by us here in the end. As long as users get notified somehow that they need to change something I'm fine with the solution.

felixfontein:
@Shaps @resmo @MarkusTeufelberger @gundalow thanks for all your feedback!

PR #9, opened 2020-03-30
cryptography backend: parse dirName, RID and otherName names

felixfontein:
##### SUMMARY
(Port of ansible/ansible#67669 to community.crypto. I'm copying its description.)

These weren't supported so far. Prompted by a [question on the ansible-project mailing list](https://groups.google.com/d/msgid/ansible-project/273386b0-ef64-4bca-902e-593b88a7f9ff%40googlegroups.com).

While adding tests, I noticed that it is pretty much impossible to use `dirName` with the pyOpenSSL backend, since the functions are designed to work with a OpenSSL config file (the syntax is `dirName:sectionName`, and `sectionName` is supposed to be a section name in that config file - not very helpful); and support for `otherName` is only partial: while it supports more complex ways of specifying it (you can specify the ASN.1 type and the value in a readable form), it can't parse the result: it will always be printed as `<unsupported>`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/module_utils/crypto.py
openssl_csr
openssl_csr_info
x509_certificate
x509_certificate_info

### DIFF
--- a/changelogs/fragments/67669-cryptography-names.yml
+++ b/changelogs/fragments/67669-cryptography-names.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- "openssl_* modules - the cryptography backend now properly supports ``dirName``, ``otherName`` and ``RID`` (Registered ID) names."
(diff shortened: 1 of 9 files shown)


felixfontein:
ready_for_review

MarkusTeufelberger:
Looks good to me

shipit

felixfontein:
@MarkusTeufelberger thanks a lot for reviewing this!

PR #12, opened 2020-03-30
Use FQCN in examples

felixfontein:
##### SUMMARY
Changes all examples to use FQCN (fully qualified collection names) instead of short names.

Mainly to make sure that people copying from the examples will end up with something that is future-proof. (I'm not aware of any plans to disable routing. But eventually we add new modules which won't have a valid shortname, and suddenly you need to use the FQCN for some modules, while you can't use it for others. So let's teach new users to use FQCNs directly :) )

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
all

### DIFF
--- a/plugins/modules/acme_account.py
+++ b/plugins/modules/acme_account.py
@@ -95,7 +95,7 @@ options:
 
 EXAMPLES = '''
 - name: Make sure account exists and has given contacts. We agree to TOS.
-  acme_account:
+  community.crypto.acme_account:
     account_key_src: /etc/pki/cert/private/account.key
     state: present
     terms_agreed: yes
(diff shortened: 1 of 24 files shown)


resmo:
I read somewhere that examples should not be adjusted to use the FQCN. @gundalow any comment on that?

felixfontein:
@resmo I've seen this question pop up before, but I've never seen a real answer to it. Would be really good to know, since I guess a lot of people will have the same question :)

gundalow:
Thanks for the ping, this made me realise that we haven't documented this yet. 

Based on discussion just now in  `#ansible-community`, I've created https://github.com/ansible-collections/overview/issues/40

Shaps:
LGTM

felixfontein:
@resmo @gundalow @Shaps thanks! :)

PR #14, opened 2020-03-31
Ansible 2.9 tests

felixfontein:
##### SUMMARY
Add CI with Ansible 2.9. Not all platforms are tested; a few should be enough to test that everything works.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/plugins/module_utils/crypto.py
+++ b/plugins/module_utils/crypto.py
@@ -424,7 +424,7 @@ def write_file(module, content, default_mode=None, path=None):
         # The path argument is only supported in Ansible 2.10+. Fall back to
         # pre-2.10 behavior of module_utils/crypto.py for older Ansible versions.
         file_args = module.load_file_common_arguments(module.params)
-        file_args['path'] = path
+        file_args['path'] = to_bytes(path)
     if file_args['mode'] is None:
         file_args['mode'] = default_mode
     # Create tempfile name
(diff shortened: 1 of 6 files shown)


plugins/module_utils/crypto.py
```diff
@@ -418,7 +418,13 @@ def write_file(module, content, default_mode=None, path=None):
     Uses file arguments from module.
     '''
     # Find out parameters for file
-    file_args = module.load_file_common_arguments(module.params, path=path)
+    try:
```
gundalow: Nice that this found a bug straight away :)

felixfontein: It's not really a bug, it's using a feature which is only available in Ansible 2.10 :)

felixfontein:
Now I finally managed to work around a (or some) bug in 2.9's ansible-test (which are already fixed in devel and should get backported soon), and to make the workaround for the Ansible 2.10 feature working.

@gundalow is it OK to have six more CI nodes for testing with Ansible 2.9?

gundalow:
> Now I finally managed to work around a (or some) bug in 2.9's ansible-test (which are already fixed in devel and should get backported soon), and to make the workaround for the Ansible 2.10 feature working.
> 
> @gundalow is it OK to have six more CI nodes for testing with Ansible 2.9?

Yes that's fine. Thanks for checking. I've made a note in the GH collections project board to review Shippable usage and Ansible-tests change detection.

felixfontein:
@gundalow thanks! :)

plugins/module_utils/crypto.py
felixfontein: (For reference, it was added in ansible/ansible#61659.)

PR #22, opened 2020-04-05
[WIP] Reproducing ansible-collections/community.general#104

felixfontein:
**DON'T MERGE THIS!***

Trying to debug strange behavior of pyOpenSSL. Potentially caused by multiple versions of the same lib lying around in the search path, or something similar (Ansible collection loader *does* do some adjustments).


felixfontein:
Interesting facts:
 1. This doesn't happen in the usual tests.
 2. It happens both with Ansible 2.9 **and** devel. So it is not caused by changes to Ansible 2.10.
 3. pyOpenSSL hasn't been released since 2019-11-18. cryptography has been released on 2020-04-02, and before that on 2019-10-17. It doesn't seem to be related to that, though: the same problem happens with cryptography 2.8.

felixfontein:
Also very interesting: this only happens when the modules are in collections (ansible/ansible#68700 replicates the same test with Ansible's stable-2.9 branch and its included openssl_privatekey module). So this could very well be related to the collection loader.

felixfontein:
Related stuff:
 1. Reproducer without collections (cannot reproduce): ansible/ansible#68700
 2. Issue in Ansible: ansible/ansible#68701

felixfontein:
I've removed some earlier tests, and added a commit to remove the hack from #36. The result should be that both nodes are unstable.

Issue #23, opened 2020-04-06
Strange error for modules importing pyOpenSSL when Ansible just upgraded cryptography on controller


felixfontein:
##### SUMMARY
See ansible/ansible#68701 and ansible-collections/community.crypto#22 for details.

This affects everything importing module_utils/crypto.py in collections.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
get_certificate
openssl_*

##### ANSIBLE VERSION
```paste below
2.9
devel
```

felixfontein:
Simplified reproducer: ansible-collections/community.crypto#22

felixfontein:
TODO: When this is fixed, re-enable consul tests in community.general (ansible-collections/community.general#117)

MarkusTeufelberger:
Maybe we should just vendor all our dependencies (at least the Python ones) in here and be done with it? That way we could also deprecate pyOpenSSL immediately...

felixfontein:
I'm not sure whether that will make our life easier, since cryptography comes with some binary components which depend on the system where it is installed. The only pure-Python dependencies are ipaddress and pyOpenSSL, and the latter depends on a new enough version of cryptography.

Anyway, this issue is about a bug in Ansible itself, which should get fixed eventually. And then this specific problem will go away. (Not all other cryptography/pyOpenSSL problems, of course :) )

MarkusTeufelberger:
Urgh, at some point I'm more and more leaning towards writing this in go or rust and statically compiling it for Linux. How's the support for binary modules in collections? Probably not even that bad I guess...

bcoca:
they work just like any other module

felixfontein:
This will be finally fixed once the routing/tombstoning PR is merged.

PR #27, opened 2020-04-10
Refactor module_utils/crypto.py

### LINKED ISSUE #16: TODO: Refactor module_utils/crypto.py
Make it a subdirectory and split that file up into several smaller files.

This allows modules to include only the parts they actually need.

Requires ~~#7 and~~ #9 to be merged first.

felixfontein:
##### SUMMARY
Expands the large monolithic `plugins/module_utils/crypto.py` into a Python package `plugins/module_utils/crypto/` with multiple smaller files.

This should not change functionality at all (except fix a tiny bug in x509_crl.py I found while changing this). CI should check essentially everything, except `ecs_certificate` (and the ECS part of `x509_certificate`) - @ctrufan can you test it?

I've used pylint and flake8 to check for errors (and clean up unnecessary imports), so the change should be pretty safe.

Fixes #16.

CC @MarkusTeufelberger @puiterwijk @Spredzy @Shaps @ctrufan

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
plugins/module_utils/crypto.py


felixfontein:
(Unstable CI is caused by #23 and unrelated to this PR.)

felixfontein:
I moved some generic file I/O methods (`load_file_if_exists` and `write_file`) into a new Python module `plugins/module_utils/io.py`. This makes them easier to use without also having to include the pyOpenSSL / cryptography feature detection.

I also added a new support function for obtaining the serial number of a `cryptography` certificate, since that changed in cryptography 1.4. Instead of repeating the same `try/except` in different modules, there's now one common utility function for that.

felixfontein:
I've added some compatibility code to `module_utils/crypto/__init__.py`. With this, even existing modules and existing code depending on `module_utils/crypto.py` should work without change - well, should, assuming ansible/ansible#68872 wouldn't be there. But once that's fixed, it will work :)

felixfontein:
(The current error in macOS is caused by pyca/cryptography#5215)

felixfontein:
@MarkusTeufelberger @puiterwijk @Spredzy @Shaps @ctrufan I would be really happy if you could review this PR :)

ctrufan:
Apologies, I've been super busy - I should have a chance to retest the ecs_certificate module some time this week. I'll see if I can find time this weekend to review the PR as well.

felixfontein:
The problem with cryptography on old macOS versions has been resolved with the release of cryptography 2.9.2. Now it's back to "unstable" as before.

MarkusTeufelberger:
Looks good to me, let's move this forward. Thanks for the work, Felix!

felixfontein:
I'll merge this now, so work can continue ;) @MarkusTeufelberger thanks for checking this! @ctrufan the change should be pretty safe for the ecs_* modules, but it probably won't hurt when you test it at some point. You should definitely test it before we release 1.0.0 of the collection anyway ;)

PR #29, opened 2020-04-11
x509_crl: support DER format for CRLs

felixfontein:
##### SUMMARY
~~Contains #27. Will be rebased and un-WIP'ed once that is merged.~~

Adds support for DER-encoded CRLs. (They seem to be more common than PEM-encoded CRLs, what is what the module currently only supports.)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
x509_crl
x509_crl_info

PR #30, opened 2020-04-11
Add x509_certificate_revocation_info module

felixfontein:
##### SUMMARY
A new module which allows to check whether a certificate is revoked, via different sources. (As suggested by @ctrufan in https://github.com/ansible/ansible/pull/67539#issuecomment-587571002) Right now, only CRLs are supported. A certificate (given by a PEM file, PEM content, or by its serial number) can be checked against a CRL file, against a CRL URL, and/or against the CRL URLs contained in the certificate (only if PEM is given, of course).

I guess support for OCSP checking will be up next, though since I'm not sure when I have time to work this, I'd be happy if this can get merged before that is added as well.

Contains #27 and #29. Will be rebased and un-WIP'ed once that is merged.

Closes ansible/ansible#53789.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
x509_certificate_revocation_info


felixfontein:
ready_for_review

plugins/modules/x509_certificate_revocation_info.py
```diff
@@ -0,0 +1,366 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2016-2017, Yanis Guenane <yanis+ansible@guenane.org>
+# Copyright: (c) 2017, Markus Teufelberger <mteufelberger+ansible@mgit.at>
```
MarkusTeufelberger, line 5: Not sure if there's much code from me in there, if anything you should at least add yourself here too. :-)

felixfontein: Haha, I think I copied that part from the wrong file :D

ctrufan:
Sorry I've been too busy to comment on much. I will note that there were some OCSP checks in my very first "add entrust support" pull request that I removed because they over-complicated the module. Wasn't tested with a variety of OCSP responders though

https://github.com/ansible/ansible/pull/59272/commits/689ddd1ea31ce36048dfc58f6f95042556afe848

plugins/modules/x509_certificate_revocation_info.py
```diff
@@ -0,0 +1,365 @@
...
+                crl = x509.load_der_x509_crl(crl_blob, default_backend())
+        except Exception as e:
+            self.module.fail_json(msg='Error while decoding CRL from {1}: {0}'.format(e, crl_source))
+
+        # Check revoced certificates
+        if 'crl_contained' not in result:
+            result['crl_contained'] = False
+            result['crl_record'] = None
+        for cert in crl:
+            if cert.serial_number == self.cert_serial_number:
```
ctrufan, line 250: In addition for serial number match, to be considered revoked a certificate also must have either:
- The same issuer as the CRL
- The "certificateIssuer" field on the CRL entry set to the same issuer as the certificate (which is one or more entries in the issuer alternate names).

Also,there's sort of a "fall through" for CRL issuer values, where if it's not set, it defaults to whatever the value was for the previous entry in the CRL. So you could have CRL Entries (don't quote me on this):
1. <CertIssuer=NotSet> Defaults to Z, which was the CRL Issuer.
2. <CertIssuer=A> Users Cert Issuer A
3. <CertIssuer=NotSet> Uses Cert Issuer A
4. <CertIssuer=NotSet> Uses Cert Issuer A
5. <CertIssuer=B> Uses Cert Issuer B
6. <CertIssuer=NotSet> Uses Cert Issuer B

Relevant section from RFC5280 https://tools.ietf.org/html/rfc5280
> 5.3.3.  Certificate Issuer
> This CRL entry extension identifies the certificate issuer associated
>    with an entry in an indirect CRL, that is, a CRL that has the
>    indirectCRL indicator set in its issuing distribution point
>    extension.  When present, the certificate issuer CRL entry extension
>    includes one or more names from the issuer field and/or issuer
>    alternative name extension of the certificate that corresponds to the
>    CRL entry.  If this extension is not present on the first entry in an
>    indirect CRL, the certificate issuer defaults to the CRL issuer.  On
>    subsequent entries in an indirect CRL, if this extension is not
>    present, the certificate issuer for the entry is the same as that for
>    the preceding entry.  This field is defined as follows:
> 
>    id-ce-certificateIssuer   OBJECT IDENTIFIER ::= { id-ce 29 }
> 
>    CertificateIssuer ::=     GeneralNames
> 
>    Conforming CRL issuers MUST include in this extension the
>    distinguished name (DN) from the issuer field of the certificate that
>    corresponds to this CRL entry.  The encoding of the DN MUST be
>    identical to the encoding used in the certificate.
> 
>    CRL issuers MUST mark this extension as critical since an
>    implementation that ignored this extension could not correctly
>    attribute CRL entries to certificates.  This specification RECOMMENDS
>    that implementations recognize this extension.

plugins/modules/x509_certificate_revocation_info.py
```diff
@@ -0,0 +1,365 @@
...
+        # Decode CRL
+        try:
+            if identify_pem_format(crl_blob):
+                crl = x509.load_pem_x509_crl(crl_blob, default_backend())
+            else:
+                crl = x509.load_der_x509_crl(crl_blob, default_backend())
+        except Exception as e:
+            self.module.fail_json(msg='Error while decoding CRL from {1}: {0}'.format(e, crl_source))
+
+        # Check revoced certificates
```
ctrufan, line 245:
```suggestion
        # Check revoked certificates
```

plugins/modules/x509_certificate_revocation_info.py
```diff
@@ -0,0 +1,365 @@
...
+            if ext is None:
+                self._report_error(soft_fail, 'No CRL Distribution Points extension found in certificate')
+            else:
+                for distribution_point in ext.value:
+                    if distribution_point.relative_name is not None:
+                        self._report_error(soft_fail, 'Distribution point with relative name found in certificate')
+                    if distribution_point.full_name is not None:
+                        had_crl_url = False
+                        for name in distribution_point.full_name:
+                            if isinstance(name, x509.UniformResourceIdentifier):
```
ctrufan, line 307: It's possible for a full name to be an LDAP entry (e.g. would show up in windows cert browser as "URL=ldap://blah", not sure about how it gets decoded by python).

plugins/modules/x509_certificate_revocation_info.py
felixfontein: Thanks!

plugins/modules/x509_certificate_revocation_info.py
felixfontein: I guess it would error out with "`ldap` is an unsupported URL schema", or something like that. Which I think is acceptable.

plugins/modules/x509_certificate_revocation_info.py
```diff
@@ -0,0 +1,365 @@
...
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = r'''
+---
+module: x509_certificate_revocation_info
+short_description: Query revocation information for X.509 certificates
```
felixfontein, line 17:
```suggestion
short_description: Query revocation information for X.509 certificates
version_added: 1.1.0
```

pabelanger:
recheck

PR #43, opened 2020-05-02
Add extra sanity tests

felixfontein:
##### SUMMARY
Idea is to add more sanity tests which ansible-test currently does not support, like:
* changelog fragment validation;
* version checking for bundled scripts (ipaddress).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
CI


felixfontein:
Hmm I think I know why it isn't working: because it's docker-in-docker already. I guess I have to do it like ansible-test and copy what is needed into the container...

felixfontein:
CC @resmo @gundalow @mattclay 

This PR is about running extra sanity checks for collections (right now community.crypto, but it could be re-used, or also moved to community.internal_test_tools as @resmo suggested). (Background: right now, ansible-test has support for drop-in sanity tests https://github.com/ansible/ansible/tree/devel/test/sanity/code-smell, but this is only allowed for ansible/ansible itself.)

These extra checks depend on the collection (like this one uses changelog linting, checking bundled versions, avoiding unwanted files in `plugins/`; community.general has nothing bundled, but has a BOTMETA.yaml file which should be validated), so there's no "one set of sanity tests which fits all collections".

I've hacked together a rool (`tests/sanity/run.py` :) ) which starts another instance (`tests/sanity/runner.py`) in a docker container (so that enough Python versions exists), installs requirements, and runs these tests. It has a nice text output, but does not (yet) create parsable data for Shippable or ansibullbot. It also has no change detection. Implementing both of them requires either using internal parts of ansible-test, copying them over, or reimplementing them. (I.e. nothing I really want to do.)

@mattclay the main difference to the code-smell tests coming with Ansible is that the tests here allow to specify a Python version used to run them and a list of requirements that should be installed ([example](https://github.com/ansible-collections/community.crypto/blob/f509c4cead29eea0d80a927fd297ee69b68624e4/tests/sanity/extra/changelog.json)), and to disable them ([example](https://github.com/ansible-collections/community.crypto/blob/f509c4cead29eea0d80a927fd297ee69b68624e4/tests/sanity/extra/update-bundled.json#L2)).

I think it would be a lot nicer if ansible-test could do this, instead of having another tool with a reduced set of features. @mattclay I know that you were thinking about having pluggable tests also for collections at some point. Do you imagine something in this direction, or something different?

I think having extra sanity checks (next to the PRs I created for ansible-test already :) ) is becoming more and more important, since more and more people start contributing to the large collections community.general and community.network.

felixfontein:
I've created a PR at community.internal_test_tools to make the runner available there, so it can also be used by other collections until ansible-test has support for plugins. I plan to deprecate it by then.

I've also adjusted this PR to use the runner from that branch.

felixfontein:
CC @Andersson007

PR #63, opened 2020-06-06
Add openssl_signature module

MarkusTeufelberger:
##### SUMMARY

Adds a new module to create or verify a Base64 encoded signature of a file.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

openssl_signature

##### ADDITIONAL INFORMATION

Originally written by @aveexy, his version is the initial commit in this PR.

Changes + tests (+ bugs/errors) are by me.

Currently still working a bit on making tests a bit more exhaustive, maybe also adding a few more features (choosing padding and hashing algorithms, allowing private keys to be used as content not only paths, allowing to verify with public keys only and not certificates, signing strings instead of files...) if they seem useful. The current state is the one that we used internally at https://mgit.at/ as part of a larger project and it seemed useful and general enough to release in this format.

Main focus of reviews should probably be also how it adheres to in-collection "standards" (parameter names, code structure), functionality wise it is not that complex fortunately.

### DIFF
--- a/plugins/modules/openssl_signature.py
+++ b/plugins/modules/openssl_signature.py
@@ -1,0 +1,346 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2016, Yanis Guenane <yanis+ansible@guenane.org>
+# Copyright: (c) 2019, Patrick Pichler <ppichler+ansible@mgit.at>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1'}
+
+DOCUMENTATION = r'''
+---
+module: openssl_signature
+short_description: Sign and verify data with openssl
+description: This module allows one to sign and verify data via certificate and private key
+requirements:
+    - Either cryptography >= 1.2.3 (older versions might work as well)
+    - Or pyOpenSSL
+author:
+    - Patrick Pichler (@aveexy)
+options:
+    action:
+        description: Action to be executed
+        type: str
+        required: true
+        choices: [ sign, verify ]
+
+    private_key:
+        description: Private key required for sign action
+        type: path
+
+    certificate:
+        description: Certificate required for verify action
+        type: path
+
+    passphrase:
+        description: Passphrase for private_key
+        type: str
+
+    path:
+        description: file to sign/verify
+        type: path
+        required: true
+
+    signature:
+        description: base64 encoded signature required for verify action
+        type: str
+
+    select_crypto_backend:
+        description:
+            - Determines which crypto backend to use.
+            - The default choice is C(auto), which tries to use C(cryptography) if available, and falls back to C(pyopenssl).
+            - If set to C(pyopenssl), will try to use the L(pyOpenSSL,https://pypi.org/project/pyOpenSSL/) library.
+            - If set to C(cryptography), will try to use the L(cryptography,https://cryptography.io/) library.
+        type: str
+        default: auto
+        choices: [ auto, cryptography, pyopenssl ]
+'''
+
+EXAMPLES = r'''
+- name: Sign example file
+  openssl_signature:
+    action: sign
+    private_key: private.key
+    path: /tmp/example_file
+  register: sig
+
+- name: Verify signature of example file
+  openssl_signature:
+    action: verify
+    certificate: cert.pem
+    path: /tmp/example_file
+    signature: sig.signature
+'''
+
+RETURN = r'''
+signature:
+    description: base64 encoded signature
+    returned: changed or success
+    type: str
+'''
+
+import os
+import traceback
+from distutils.version import LooseVersion
+import base64
+
+MINIMAL_PYOPENSSL_VERSION = '0.6'
+MINIMAL_CRYPTOGRAPHY_VERSION = '1.2.3'
+
+PYOPENSSL_IMP_ERR = None
+try:
+    import OpenSSL
+    from OpenSSL import crypto
+
+    PYOPENSSL_VERSION = LooseVersion(OpenSSL.__version__)
+except ImportError:
+    PYOPENSSL_IMP_ERR = traceback.format_exc()
+    PYOPENSSL_FOUND = False
+else:
+    PYOPENSSL_FOUND = True
+
+CRYPTOGRAPHY_IMP_ERR = None
+try:
+    import cryptography
+    import cryptography.hazmat.primitives.asymmetric.rsa
+    import cryptography.hazmat.primitives.asymmetric.dsa
+    import cryptography.hazmat.primitives.asymmetric.ec
+    import cryptography.hazmat.primitives.asymmetric.padding
+    import cryptography.hazmat.primitives.hashes
+
+    CRYPTOGRAPHY_VERSION = LooseVersion(cryptography.__version__)
+except ImportError:
+    CRYPTOGRAPHY_IMP_ERR = traceback.format_exc()
+    CRYPTOGRAPHY_FOUND = False
+else:
+    CRYPTOGRAPHY_FOUND = True
+
+from ansible.module_utils import crypto as crypto_utils
+from ansible.module_utils._text import to_native, to_bytes
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+
+
+class SignatureBase(crypto_utils.OpenSSLObject):
+
+    def __init__(self, module, backend):
+        super(SignatureBase, self).__init__(
+            module.params['path'],
+            'present',
+            False,
+            module.check_mode
+        )
+
+        self.backend = backend
+
+        self.action = module.params['action']
+        self.signature = module.params['signature']
+        self.passphrase = module.params['passphrase']
+        self.private_key = module.params['private_key']
+        self.certificate = module.params['certificate']
+
+    def generate(self):
+        # Empty method because crypto_utils.OpenSSLObject wants this
+        pass
+
+    def dump(self):
+        # Empty method because crypto_utils.OpenSSLObject wants this
+        pass
+
+
+# Implementation with using pyOpenSSL
+class SignaturePyOpenSSL(SignatureBase):
+
+    def __init__(self, module, backend):
+        super(SignaturePyOpenSSL, self).__init__(module, backend)
+
+    def run(self):
+        try:
+            result = dict()
+
+            with open(self.path, "rb") as f:
+                _in = f.read()
+
+            if self.action == "verify":
+                _signature = base64.b64decode(self.signature)
+                certificate = crypto_utils.load_certificate(self.certificate, backend=self.backend)
+
+                try:
+                    OpenSSL.crypto.verify(certificate, _signature, _in, 'sha256')
+                except Exception:
+                    self.module.fail_json(
+                        msg="Verification failed"
+                    )
+
+            elif self.action == "sign":
+                private_key = crypto_utils.load_privatekey(
+                    self.private_key,
+                    None if self.passphrase is None else to_bytes(self.passphrase),
+                    backend=self.backend
+                )
+
+                out = OpenSSL.crypto.sign(private_key, _in, "sha256")
+                result['signature'] = base64.b64encode(out)
+
+            return result
+        except Exception as e:
+            raise crypto_utils.OpenSSLObjectError(e)
+
+
+# Implementation with using cryptography
+class SignatureCryptography(SignatureBase):
+
+    def __init__(self, module, backend):
+        super(SignatureCryptography, self).__init__(module, backend)
+
+    def run(self):
+        _padding = cryptography.hazmat.primitives.asymmetric.padding.PKCS1v15()
+        _hash = cryptography.hazmat.primitives.hashes.SHA256()
+
+        result = dict()
+
+        try:
+            with open(self.path, "rb") as f:
+                _in = f.read()
+
+            if self.action == "verify":
+                _signature = base64.b64decode(self.signature)
+                public_key = crypto_utils.load_certificate(self.certificate, backend=self.backend).public_key()
+
+                try:
+                    if isinstance(public_key, cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey):
+                        public_key.verify(_signature, _in, _padding, _hash)
+
+                    elif isinstance(public_key, cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey):
+                        public_key.verify(_signature, _in, cryptography.hazmat.primitives.asymmetric.ec.ECDSA(_hash))
+
+                    elif (isinstance(public_key, cryptography.hazmat.primitives.asymmetric.ed25519.Ed25519PublicKey) or
+                          isinstance(public_key, cryptography.hazmat.primitives.asymmetric.ed448.Ed448PublicKey)):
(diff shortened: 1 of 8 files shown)


MarkusTeufelberger:
And so it begins...

FreeBSD apparently uses LibreSSL which doesn't support `ed25519` or `ed448`.

OpenSSL in their eternal wisdom decided that `openssl dgst` shouldn't work for `ed25519` or `ed448`, so I now have to hope that `openssl pkeyutl` works on all platforms... which of course it doesn't.

felixfontein:
Out of curiosity: why add a pyOpenSSL backend at all, if you plan to remove it "soon"?

MarkusTeufelberger:
It would be the only one without pyopenssl and it wouldn't run at all on Ubuntu Xenial for example (https://packages.ubuntu.com/de/xenial-updates/python-cryptography). The integration tests only pass there, because they don't actually do anything.

MarkusTeufelberger:
Urgh, `openssl pkeyutl -help` seems to return `1` on some platforms.

MarkusTeufelberger:
Otoh OpenSSL in OSX seems to return 0, even if you ask for help for a command it doesn't understand and where it also prints out "openssl:Error: 'pkeyutl' is an invalid command."...

felixfontein:
The x509_crl and x509_crl_info modules only have a cryptography backend, so it definitely wouldn't be the only one. But if pyOpenSSL is the way to get it to work on a platform you need to support, feel free to support it :)

MarkusTeufelberger:
The bigger pain point at the moment is anyways OpenSSL itself... I'm close to writing some feature detection for that whole mess, but I'm afraid that that would again mean a lot of edge cases. There aren't even that many distros tested in shippable! :-/

MarkusTeufelberger:
Hm, pyopenssl (and OpenSSL too) seems to have issues in general with Ed25519 and Ed448 signatures.

MarkusTeufelberger:
FreeBSD 11 manages to die quite spectacularly when signing with `secp256r1` keys... I'll just disable these tests for now on FreeBSD and move on, there needs to be some rework in 1-2 years when we remove support for pyopenssl anyways.

MarkusTeufelberger:
Alright, this is now ready for review. I'll squash down my commits before merging.

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+      was deprecated in Ansible 2.9 and will be removed in Ansible 2.13.
+requirements:
+    - Either cryptography >= 1.4 (some key types require newer versions)
+    - Or pyOpenSSL >= 0.11 (Ed25519 and Ed448 keys are not supported with this backend)
+author:
+    - Patrick Pichler (@aveexy)
+    - Markus Teufelberger (@MarkusTeufelberger)
+options:
+    action:
+        description: Action to be executed
```
felixfontein, line 29:
```suggestion
        description: Action to be executed.
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: openssl_signature
+short_description: Sign and verify data with openssl
+description:
+    - This module allows one to sign and verify data via a certificate and a private key
```
felixfontein, line 16:
```suggestion
    - This module allows one to sign and verify data via a certificate and a private key.
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+    - Markus Teufelberger (@MarkusTeufelberger)
+options:
+    action:
+        description: Action to be executed
+        type: str
+        required: true
+        choices: [ sign, verify ]
+    privatekey_path:
+        description:
+            - The path to the private key to use when signing
```
felixfontein, line 35:
```suggestion
            - The path to the private key to use when signing.
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+        type: path
+    privatekey_passphrase:
+        description:
+            - The passphrase for the private key.
+            - This is required if the private key is password protected.
+        type: str
+    path:
+        description: File to sign/verify
```
felixfontein, line 43:
```suggestion
        description: File to sign/verify.
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+        type: path
+        required: true
+    certificate:
+        description: Certificate required for verify action
```
felixfontein, line 47:
```suggestion
        description: Certificate required for verify action.
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+        type: path
+    signature:
+        description: Base64 encoded signature required for verify action
```
felixfontein, line 50:
```suggestion
        description: Base64 encoded signature required for verify action.
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+  openssl_signature:
+    action: verify
+    certificate: cert.pem
+    path: /tmp/example_file
+    signature: "{{ sig.signature }}"
+'''
+
+RETURN = r'''
+signature:
+    description: Base64 encoded signature
```
felixfontein, line 90:
```suggestion
    description: Base64 encoded signature.
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+    returned: changed or success
```
felixfontein, line 91:
```suggestion
    returned: success
```
Changed is part of success :)

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+                                  "cryptography (>= {0}) or PyOpenSSL (>= {1})").format(
+                MINIMAL_CRYPTOGRAPHY_VERSION,
+                MINIMAL_PYOPENSSL_VERSION))
+    try:
+        if backend == 'pyopenssl':
+            if not PYOPENSSL_FOUND:
+                module.fail_json(msg=missing_required_lib('pyOpenSSL >= {0}'.format(MINIMAL_PYOPENSSL_VERSION)),
+                                 exception=PYOPENSSL_IMP_ERR)
+            module.deprecate('The module is using the PyOpenSSL backend. This backend has been deprecated',
+                             version='2.13')
```
felixfontein, line 367:
```suggestion
                             version='2.0.0', collection_name='community.crypto')
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+      DSA and ECDSA keys: C(cryptography) >= 1.5
+      ed448 and ed25519 keys: C(cryptography) >= 2.6
+seealso:
+    - module: x509_certificate
+    - module: openssl_privatekey
+'''
+
+EXAMPLES = r'''
+- name: Sign example file
+  openssl_signature:
```
felixfontein, line 74:
```suggestion
  community.crypto.openssl_signature:
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+    action: sign
+    privatekey_path: private.key
+    path: /tmp/example_file
+  register: sig
+
+- name: Verify signature of example file
+  openssl_signature:
```
felixfontein, line 81:
```suggestion
  community.crypto.openssl_signature:
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+    - Or pyOpenSSL >= 0.11 (Ed25519 and Ed448 keys are not supported with this backend)
+author:
+    - Patrick Pichler (@aveexy)
+    - Markus Teufelberger (@MarkusTeufelberger)
+options:
+    action:
+        description: Action to be executed
+        type: str
+        required: true
+        choices: [ sign, verify ]
```
felixfontein, line 32: Conceptually, I think it would be better to have two modules, one for signing, the other for verifying. I think the idea of having modules which do different actions depending on a module option is nowadays discouraged in Ansible.

How about having something like `openssl_signature_sign` and `openssl_signature_verify`?

MarkusTeufelberger: I'm not opposed, though the content to boilerplate/docs ratio there would be even higher.

Instead of `openssl_signature_verify` there could also be an `openssl_signature_info` module that tells you if a signature is valid/verified? That would follow the current pattern here maybe even more...

felixfontein: Yep, it definitely would increase the content/boilerplate ratio.

`_info` is probably better. I first thought about that too, and was wondering whether it is a good fit and first decided against it, but now (after thinking some more about it) I guess it's better.

MarkusTeufelberger: Alright, I'll go the `openssl_signature`/`openssl_signature_info` route. Time to train those Ctrl + c / Ctrl + v skills! :-D

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,383 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: openssl_signature
+short_description: Sign and verify data with openssl
```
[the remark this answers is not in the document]
felixfontein: Should be 1.1.0 now :)

MarkusTeufelberger:
rebased + rewrote to 2 different modules. The tests are identical for now, not sure if there's much point in separating them?

Things that would be nice, but could be implemented later as additional features:

`openssl_signature` - choose hashing algorithm and padding algorithm used when signing (big can of worms when it comes to which key type allows which settings though)
`openssl_signature_info` - more return values than just `changed: false` and `valid: true/false`, e.g. hashing algo and passing used (not sure how to reliably detect that though)
`openssl_signature_info` - also allow public keys rather than parsing them out of a certificate (should be fairly straightforward)

felixfontein:
> The tests are identical for now, not sure if there's much point in separating them?

No; I would only keep one copy, and add the other module in `aliases`, like https://github.com/ansible-collections/community.crypto/blob/main/tests/integration/targets/x509_crl/aliases#L2 (the x509_crl tests also test x509_crl_info).

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,321 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: openssl_signature
+short_description: Sign data with openssl
```
felixfontein, line 14:
```suggestion
short_description: Sign data with openssl
version_added: 1.1.0
```

plugins/modules/openssl_signature_info.py
```diff
@@ -0,0 +1,353 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: openssl_signature_info
+short_description: Verify signatures with openssl
```
felixfontein, line 14:
```suggestion
short_description: Verify signatures with openssl
version_added: 1.1.0
```

plugins/modules/openssl_signature.py
```diff
@@ -0,0 +1,321 @@
...
+        choices: [ auto, cryptography, pyopenssl ]
+notes:
+    - |
+      When using the C(cryptography) backend, the following key types require at least the following C(cryptography) version:
+      RSA keys: C(cryptography) >= 1.4
+      DSA and ECDSA keys: C(cryptography) >= 1.5
+      ed448 and ed25519 keys: C(cryptography) >= 2.6
+seealso:
+    - module: openssl_signature_info
+    - module: openssl_privatekey
```
felixfontein, line 66:
```suggestion
    - module: community.crypto.openssl_signature_info
    - module: community.crypto.openssl_privatekey
```

plugins/modules/openssl_signature_info.py
```diff
@@ -0,0 +1,353 @@
...
+        choices: [ auto, cryptography, pyopenssl ]
+notes:
+    - |
+      When using the C(cryptography) backend, the following key types require at least the following C(cryptography) version:
+      RSA keys: C(cryptography) >= 1.4
+      DSA and ECDSA keys: C(cryptography) >= 1.5
+      ed448 and ed25519 keys: C(cryptography) >= 2.6
+seealso:
+    - module: openssl_signature
+    - module: x509_certificate
```
felixfontein, line 65:
```suggestion
    - module: community.crypto.openssl_signature
    - module: community.crypto.x509_certificate
```

plugins/modules/openssl_signature_info.py
```diff
@@ -0,0 +1,353 @@
...
+
+- name: Make sure the signature is valid
+  assert:
+    that:
+      - verify.valid
+'''
+
+RETURN = r'''
+valid:
+    description: True means the signature was valid for the given file, False means it wasn't.
```
felixfontein, line 90:
```suggestion
    description: C(true) means the signature was valid for the given file, C(false) means it wasn't.
```
Jinja2 booleans are lower case.

tests/integration/targets/openssl_signature/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+    provider: selfsigned
+    path: '{{ output_dir }}/{{item.backend}}_certificate_{{ item.type }}_{{ item.passwd }}.pem'
+    privatekey_path: '{{ output_dir }}/{{item.backend}}_privatekey_{{ item.type }}_{{ item.passwd }}.pem'
+    privatekey_passphrase: '{{ item.privatekey_passphrase | default(omit) }}'
+    csr_path: '{{ output_dir }}/{{item.backend}}_{{ item.type }}_{{ item.passwd }}.csr'
+  loop: '{{ all_tests }}'
+
+- name: Create statement to be signed
+  copy:
+    content: "Erst wenn der Subwoofer die Katze inhaliert, fickt der Bass richtig übel. -- W.A. Mozart"
```
felixfontein, line 103:
```suggestion
    content: "Erst wenn der Subwoofer die Katze inhaliert, fickt der Bass richtig übel. -- W.A. Mozart"
```
:D

MarkusTeufelberger:
Still needs a squash, hoped to do it this weekend, but alas... I'm still on it though and will merge then.

felixfontein:
@MarkusTeufelberger the GitHub merge configured for this repo will already squash all commits :)

MarkusTeufelberger:
Yeah, but it will likely preserve all these ugly commit messages...

felixfontein:
True :)

felixfontein:
I've squashed the commits into two commits (one per author). I'll rebase-merge once the tests pass.

felixfontein:
Gna. Why does GitHub always changes the merge button selection when reloading?! This is exactly what I wanted to prevent...

felixfontein:
With @gundalow's help I purged the commit from `main` and then manually pushed the two commits from this PR to `main`. So now we have two commits for this PR, one per author, which both mention this PR in their commit message.

felixfontein:
@MarkusTeufelberger and Patrick Pichler, thanks a lot for contributing these new modules! :)

PR #67, opened 2020-06-14
Porting guide

felixfontein:
##### SUMMARY
Add porting guide entries from ansible/ansible's 2.10 porting guide as changelog fragments.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
changelog

### DIFF
--- a/changelogs/fragments/letsencrypt.yml
+++ b/changelogs/fragments/letsencrypt.yml
@@ -1,0 +1,2 @@
+removed_features:
+- "The ``letsencrypt`` module has been removed. Use ``acme_certificate`` instead."
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Community.crypt specific things
+/changelogs/.plugin-cache.yaml
+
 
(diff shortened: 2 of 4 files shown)


felixfontein:
CC @acozine @samccann

changelogs/fragments/letsencrypt.yml
```diff
@@ -0,0 +1,2 @@
+removed_features:
+- "The ``letsencrypt`` module has been removed. Use ``acme_certificate`` instead."
```
samccann, line 2: Let's talk about this on #ansible-doc - we used to have these modules as links. @abadger - would it work if this was changed to R(acme_certificate_module) or something like that?

felixfontein: This is for the changelog, it needs to use pure reStructuredText. When included in the docsite RST refs would work, but the changelog is (I think) often viewed as a text file on github (i.e. with basic RST formatting) - and then they won't work.

felixfontein:
@samccann thanks for reviewing this!

PR #75, opened 2020-06-20
Prepare release

felixfontein:
##### SUMMARY
Do some generic cleanup; set deprecation versions to 2.0.0.

Fixes #73.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
collection

PR #78, opened 2020-07-01
x509_certificate: Update docs for macOS 10.15 requirements

Akasurde:
##### SUMMARY

selfsigned_not_after and ownca_not_after values for macOS 10.15,
can be +825d. Updated the docs accordingly.

Migrated from https://github.com/ansible/ansible/pull/64563

Thanks to Sironheart

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
changelogs/fragments/64563_openssl_cert_mac_os.yml
plugins/modules/x509_certificate.py

### DIFF
--- a/plugins/modules/x509_certificate.py
+++ b/plugins/modules/x509_certificate.py
@@ -155,6 +155,8 @@ options:
             - Note that if using relative time this module is NOT idempotent.
             - If this value is not specified, the certificate will stop being valid 10 years from now.
             - This is only used by the C(selfsigned) provider.
+            - On macOS 10.15 and onwards, TLS server certificates must have a validity period of 825 days or fewer.
+              Please see U(https://support.apple.com/en-us/HT210176) for more details.
         type: str
         default: +3650d
         aliases: [ selfsigned_notAfter ]
@@ -245,6 +247,8 @@ options:
             - Note that if using relative time this module is NOT idempotent.
             - If this value is not specified, the certificate will stop being valid 10 years from now.
             - This is only used by the C(ownca) provider.
+            - On macOS 10.15 and onwards, TLS server certificates must have a validity period of 825 days or fewer.
+              Please see U(https://support.apple.com/en-us/HT210176) for more details.
         type: str
         default: +3650d
 


changelogs/fragments/64563_openssl_cert_mac_os.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- x509_certificate - add a note about macOS 10.15 requirements (https://github.com/ansible/ansible/pull/64563).
```
felixfontein, line 2: Docs only changes don't get changelog entries.

MarkusTeufelberger:
Still a weak no from my side as this changes defaults (though not in the code, so there will likely be some documentation/code mismatches highlighted in CI).

If you need a certificate for a browser, you'll likely use a proper CA anyways, not a selfsigned/-managed certificate. Also this limitation is not exclusive to macOS (and actually only refers to Apple's browser software, not the OS?) + a few weeks ago they also announced that they will now reduce this time further to 1 year validity (+ some grace period).

MarkusTeufelberger:
Oh, apparently I looked in the wrong place? I could swear there was also a change in the default value...

Akasurde:
> Oh, apparently I looked in the wrong place? I could swear there was also a change in the default value...

Yes, this is just doc change. https://github.com/ansible/ansible/pull/64563 changed default values.

MarkusTeufelberger:
https://support.apple.com/en-us/HT211025 looks like it is only affecting trusted CAs, Google already also has communicated a similar policy for Chrome, so this will happen (Code here: https://source.chromium.org/chromium/chromium/src/+/master:net/cert/cert_verify_proc.cc;l=978;drc=master?originalUrl=%2F)

felixfontein:
What we could do is remove the defaults for 2.0.0, and deprecate them for 1.0.0. Then people always have to explicitly say what they want, and are not stuck with a "bad" default.

felixfontein:
(BTW: the ansible/third-party-check failure is unrelated to this PR, and will be fixed by #75.)

felixfontein:
Anyway, if we do the deprecation, I'd do it in a follow-up. Should we update this to also mention Chrome (would probably need a better link), or just merge the current version?

MarkusTeufelberger:
The limitation to ~1 year is only for actual widely trusted certificates by professional CAs, they just won't give/sell you a certificate valid for 10 years anyways. The 825 days limitation in macOS is apparently also for selfsigned/homegrown stuff which is likely about hurting/limiting MITM-as-a-service vendors.

I'm fine with a note about this limitation in macOS (though it would be interesting if similar things exist on e.g. Windows?). A simple workaround even on macOS would be to just retroactively issue the long certificate from a date before 2019-07-01 by the way... ;-)

felixfontein:
Restarting tests.

felixfontein:
I guess the third-party-check does not automatically rebase before running. Let's simply ignore it...

felixfontein:
Hmm, and now it is fine? Whatever.

@MarkusTeufelberger do you mind if I merge?

MarkusTeufelberger:
Not at all, let's ship it - after all it could lead to a little less confusion why something isn't accepted on some OS.

felixfontein:
@Akasurde thanks for fixing this! @MarkusTeufelberger thanks for reviewing!

Akasurde:
@MarkusTeufelberger @felixfontein Thanks for reviewing.

Issue #76, opened 2020-06-21
Lack of documentation in new recommended way of generating openssl certificates


Akasurde:
_From @alexcernat on Jun 20, 2020 10:23_

##### SUMMARY

I am using this code to generate self signed certificates:

```
- name: Verify SSL certificate
  openssl_certificate:
    provider: assertonly
    path: "{{ path_to_crt }}"
    state: present
    issuer:
      CN: "common name"
      O: "organization"
    subject:
      CN: "{{ ansible_fqdn }}"
    valid_in: "2592000" # 30 days
  delegate_to: localhost
  ignore_errors: True
  register: verify_cert

- name: Generate SSL certificate
  my_ownca_generate:
    common_name: "{{ ansible_fqdn }}"
    key_path: "{{ path_to_key }}"
    cert_path: "{{ path_to_crt }}"
  delegate_to: localhost
  when: verify_cert.failed

```
First step is checking the actual certificate, and if it doesn't exists or is not valid (CA checks, CN checks, validity etc.), then the certificate is generated in step 2.  The fact that I am using my own module to generate the certificate is irrelevant, the fact is that I need to run step 2 (generation) only if step 1 (checking) fails (or maybe if you have a better approach ...?)

From ansible 2.9 using openssl_certificate to check the validity of a certificate is deprecated (and set to be removed in 2.13 IIRC), the documentation suggest that I should use openssl_certificate_info to check different certificate parameters, but only an "assert" method is presented in the docs.

How can I translate that "when: verify_cert.failed" from step 2 in order to work with the new openssl_certificate_info module ? I think that such an example should be provided in the docs.

Also, as a feature request, I believe that a "serial" parameter should be included in the ownca module, or at least increase somehow the "randomness" of that serial number. Last time I've check the code, IIRC the serial was between 1 and 65535, which is not quite so "random".

##### ISSUE TYPE
- Documentation Report

##### ANSIBLE VERSION

```paste below
ansible 2.9.9
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.16 (default, Oct 10 2019, 22:02:15) [GCC 8.3.0]
```

##### CONFIGURATION

```paste below

```

##### ADDITIONAL INFORMATION

[...]

PR #90, opened 2020-07-20
Increase # of bits for random serial numbers of certificates with PyOpenSSL backend

felixfontein:
##### SUMMARY
Fixes part of #76. Instead of using a random number betwen 1000 and 99999 (both inclusive), selects a 160 bit random number between 2^160 and 2^161 - 1 (both inclusive). No special random input source is used, so the quality is not very high. I think that's ok for serial numbers of certificates (and already a lot better than what we had before).

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
openssl_certificate

### DIFF
--- a/changelogs/fragments/90-openssl_certificate-pyopenssl-serial.yml
+++ b/changelogs/fragments/90-openssl_certificate-pyopenssl-serial.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- "openssl_certificate - the PyOpenSSL backend now uses 160 bits of randomness for serial numbers, instead of a random number between 1000 and 99999. Please note that this is not a high quality random number (https://github.com/ansible-collections/community.crypto/issues/76)."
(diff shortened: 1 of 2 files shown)


felixfontein:
ready_for_review

MarkusTeufelberger:
LGTM, though I wonder why one would set the lower limit at 2^160 instead of 0 for serial numbers? Seems to cut the whole possible range in half...

felixfontein:
It makes sure that the bitlength is always 161 bits ;-) It's more to make it similar to the old code (which never went below 1000).

I don't mind removing that part. The chance that this is actually 0 is pretty low anyway.

felixfontein:
(Also always setting bit 160 is less bad as always setting bit 0 IMO, which is another common choice to make sure the serial cannot be 0 ;) )

felixfontein:
I've adjusted the algorithm to return a <= 160bit number between 1000 and 2^160-1.

felixfontein:
@MarkusTeufelberger thanks for reviewing this!

PR #94, opened 2020-07-27
Tests

felixfontein:
##### SUMMARY
Fixes test aliases so that:
 1. all tests (except ACME, since `cloud/acme` does not work there) are run with the OS-specific containers;
 2. almost all tests are run with the "cloud" containers, which contain a minimal setup, but cover all supported Python versions from 2.6 to 3.9.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/shippable.yml
+++ b/shippable.yml
@@ -57,7 +57,7 @@ matrix:
     - env: T=2.9/units/3.8/1
     - env: T=2.9/rhel/7.8/1
     - env: T=2.9/linux/ubuntu1804/1
-    - env: T=2.9/cloud/3.6/1
+    - env: T=2.9/cloud/3.5/1
 
 branches:
   except:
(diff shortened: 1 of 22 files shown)


felixfontein:
Looks like this currently does not work with ansible-test. The problem is that when the `--docker default` container is used, Python dependencies **must** be installed with `pip` and not with `package` (or one of the OS-specific package managers), to avoid changing the default OS Python, and installing the packages for the wrong Python version. Unfortunately it is not possible (without really ugly hacks, like checking for the distribution and comparing the Python version to the expected default Python version - I really don't want to maintain that) to detect whether the tests are running in the `default` container or not.

I've reverted these parts and will merge once tests pass, since the remaining small changes should still go in.

PR #95, opened 2020-07-30
Add support for PKCS#11 tokens to openssh_cert.

s-hamann:
##### SUMMARY
This PR adds the parameter pkcs11_provider, which can be set to the name of or path to a PKCS#11 library (e.g. libpkcs11.so). ssh-keygen will then use this library to have the token make any required signatures.
If this is used, signing_key needs to be set to a file containing the public key that matches the private key on the token.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
`pkcs11_provider` is simply passed to `ssh-keygen`. Usual choices include `libpkcs11.so`, `libykcs11.so` (for YubiKey) and possibly others. I think it's a bit awkward to pass a library in order to use a hardware token, but that's how `ssh-keygen` handles it and I do not think we could reasonably provider a better user interface.

If the token requires a PIN to use the key, things may get a little complicated.
If `ssh-keygen` can not get the PIN from a TTY, it uses ssh-askpass, which should be OK for interactive use. However, I could not get it to run for some reason.
When using a local connection instead of SSH, the module/`ssh-keygen` interactively asks for the PIN on the terminal, which should be sufficient for the case where the PKCS#11 token is connected to the Ansible controller, not to the target system(s). I believe that to be the common use case.
I did not find a way of providing the PIN via Ansible, unfortunately.

Note: I did not add any tests, since that would require an PKCS#11 token. It may be possible to emulate that somehow, but I did not look into this topic.

### DIFF
--- a/changelogs/fragments/openssh_cert-pkcs11.yml
+++ b/changelogs/fragments/openssh_cert-pkcs11.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "openssh_cert - add support for PKCS#11 tokens (https://github.com/ansible-collections/community.crypto/pull/95)."
--- a/plugins/modules/openssh_cert.py
+++ b/plugins/modules/openssh_cert.py
@@ -43,8 +43,16 @@ options:
     signing_key:
         description:
             - The path to the private openssh key that is used for signing the public key in order to generate the certificate.
+            - If the private key is on a PKCS#11 token (I(pkcs11_provider)), set this to the path to the public key instead.
             - Required if I(state) is C(present).
         type: path
+    pkcs11_provider:
+        description:
+            - To use a signing key that resides on a PKCS#11 token, set this to the name (or full path) of the shared library to use with the token.
+              Usually C(libpkcs11.so).
+            - If this is set, I(signing_key) needs to point to a file containing the public key of the CA.
+        type: str
+        version_added: 1.1.0
     public_key:
         description:
             - The path to the public key that will be signed with the signing key in order to generate the certificate.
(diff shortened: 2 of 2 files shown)


changelogs/fragments/openssh_cert-pkcs11.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - openssh_cert - add support for PKCS#11 tokens
```
felixfontein, line 2:
```suggestion
  - "openssh_cert - add support for PKCS#11 tokens (https://github.com/ansible-collections/community.crypto/pull/95)."
```
The quotes are probably necessary since `#` starts comments in YAML files.

plugins/modules/openssh_cert.py
```diff
@@ -43,8 +43,15 @@
...
             - The path to the private openssh key that is used for signing the public key in order to generate the certificate.
+            - If the private key is on a PKCS#11 token (I(pkcs11_provider)), set this to the path to the public key instead.
             - Required if I(state) is C(present).
         type: path
+    pkcs11_provider:
+        description:
+            - To use a signing key that resides on a PKCS#11 token, set this to the name (or full path) of the shared library to use with the token.
+              Usually C(libpkcs11.so).
+            - If this is set, I(signing_key) needs to point to a file containing the public key of the CA.
+        type: str
```
felixfontein, line 54:
```suggestion
        type: str
        version_added: 1.1.0
```

s-hamann:
Thank you for your feedback.
Yes, I manually tested this and successfully signed host certificates using a YubiKey. I only tried the specific set of parameters that I need right now, not all possible variations. However, since this PR only adds a parameter that is passed to `ssh-keygen` as is, I believe that should be OK.

changelogs/fragments/openssh_cert-pkcs11.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "openssh_cert - add support for PKCS#11 tokens (https://github.com/ansible-collections/community.crypto/pull/95)"
```
felixfontein, line 2:
```suggestion
  - "openssh_cert - add support for PKCS#11 tokens (https://github.com/ansible-collections/community.crypto/pull/95)."
```

s-hamann: Sorry, I haven't been paying attention. Fixed.

felixfontein:
@s-hamann thanks a lot for the contribution!

PR #102, opened 2020-08-15
acme_certificate - allow to select first certificate in chain

felixfontein:
##### SUMMARY
When testing letsencrypt/pebble#318 I found that in the tests, the *first* (as opposed to the last) certificate in the chain needs to be matched in some cases. This adds a selector `first`, and adjusts the test to use this selector.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
acme_certificate


felixfontein:
CC @resmo

felixfontein:
@resmo thanks for reviewing this!

PR #117, opened 2020-10-05
openssh_cert: Implement use_agent option to get signing key from ssh-agent

### LINKED ISSUE #116: openssh_cert: Make it possible to use a signing key stored in ssh-agent
##### SUMMARY
If the ssh signing key is password protected, and the user has it unlocked in an ssh-agent, an additional flag needs to be passed to ssh-keygen in order for it to use the key from the agent when signing. I propose an additional optional module boolean argument to add the flag needed.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
openssh_cert.py

##### ADDITIONAL INFORMATION
See summary.

```yaml
- openssh_cert:
    type: user
    signing_key: /path/to/private_key
    use_agent: yes
    public_key: /path/to/public_key.pub
    path: /path/to/certificate
    valid_from: always
    valid_to: forever
```

dougstanley:
##### SUMMARY

Implements an additional optional boolean argument to the openssh_cert module called `use_agent` which simply passes the additional `-U` flag to `ssh-keygen` to tell it to look for the signing key in the ssh-agent.

Fixes #116 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

openssh_cert

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/117-openssh_cert-use-ssh-agent.yml
+++ b/changelogs/fragments/117-openssh_cert-use-ssh-agent.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- openssh_cert - add module parameter ``use_agent`` to enable using signing keys stored in ssh-agent (https://github.com/ansible-collections/community.crypto/issues/116).
--- a/plugins/modules/openssh_cert.py
+++ b/plugins/modules/openssh_cert.py
@@ -53,6 +53,12 @@ options:
             - If this is set, I(signing_key) needs to point to a file containing the public key of the CA.
         type: str
         version_added: 1.1.0
+    use_agent:
+        description:
+            - Should the ssh-keygen use a CA key residing in a ssh-agent.
+        type: bool
+        default: false
+        version_added: 1.3.0
     public_key:
         description:
             - The path to the public key that will be signed with the signing key in order to generate the certificate.
@@ -216,12 +222,14 @@ import tempfile
 
 from datetime import datetime
 from datetime import MINYEAR, MAXYEAR
+from distutils.version import LooseVersion
 from shutil import copy2, rmtree
 
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils._text import to_native
 
 from ansible_collections.community.crypto.plugins.module_utils.crypto.support import convert_relative_to_datetime
+from ansible_collections.community.crypto.plugins.module_utils.crypto.openssh import parse_openssh_version
 
 
 class CertificateError(Exception):
@@ -235,6 +243,7 @@ class Certificate(object):
         self.force = module.params['force']
         self.type = module.params['type']
         self.signing_key = module.params['signing_key']
+        self.use_agent = module.params['use_agent']
         self.pkcs11_provider = module.params['pkcs11_provider']
         self.public_key = module.params['public_key']
         self.path = module.params['path']
@@ -272,6 +281,9 @@ class Certificate(object):
 
             if self.pkcs11_provider:
                 args.extend(['-D', self.pkcs11_provider])
+
+            if self.use_agent:
+                args.extend(['-U'])
 
             validity = ""
 
@@ -547,6 +559,7 @@ def main():
             force=dict(type='bool', default=False),
             type=dict(type='str', choices=['host', 'user']),
             signing_key=dict(type='path'),
+            use_agent=dict(type='bool', default=False),
             pkcs11_provider=dict(type='str'),
             public_key=dict(type='path'),
             path=dict(type='path', required=True),
@@ -562,6 +575,21 @@ def main():
         add_file_common_args=True,
         required_if=[('state', 'present', ['type', 'signing_key', 'public_key', 'valid_from', 'valid_to'])],
     )
+
+    if module.params['use_agent']:
+        ssh = module.get_bin_path('ssh', True)
+        proc = module.run_command([ssh, '-Vq'])
+        ssh_version_string = proc[2].strip()
+        ssh_version = parse_openssh_version(ssh_version_string)
+        if ssh_version is None:
+            module.fail_json(msg="Failed to parse ssh version")
+        elif LooseVersion(ssh_version) < LooseVersion("7.6"):
+            module.fail_json(
+                msg=(
+                    "Signing with CA key in ssh agent requires ssh 7.6 or newer."
+                    " Your version is: %s"
+                ) % ssh_version_string
+            )
 
     def isBaseDir(path):
         base_dir = os.path.dirname(path) or '.'
--- a/tests/integration/targets/openssh_cert/tasks/main.yml
+++ b/tests/integration/targets/openssh_cert/tasks/main.yml
@@ -411,3 +411,83 @@
       path: '{{ output_dir }}/id_key'
       state: absent
     check_mode: yes
+
+- name: openssh_cert integration tests that require ssh-agent
+  when: openssh_version is version("7.6",">=")
+  environment:
+    SSH_AUTH_SOCK: "{{ openssh_agent_sock }}"
+  block:
+  - name: Generate keypair for agent tests
+    openssh_keypair:
+      path: '{{ output_dir }}/id_key'
+      type: rsa
+  - name: Generate always valid cert using agent without key in agent (should fail)
+    openssh_cert:
+      type: user
+      signing_key: '{{ output_dir }}/id_key'
+      public_key: '{{ output_dir }}/id_key.pub'
+      path: '{{ output_dir }}/id_cert_with_agent'
+      use_agent: yes
+      valid_from: always
+      valid_to: forever
+    register: rc_no_key_in_agent
+    ignore_errors: yes
+  - name: Make sure cert creation with agent fails if key not in agent
+    assert:
+      that:
+        - rc_no_key_in_agent is failed
+        - "'agent contains no identities' in rc_no_key_in_agent.msg or 'not found in agent' in rc_no_key_in_agent.msg"
+  - name: Add key to agent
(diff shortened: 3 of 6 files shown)


MarkusTeufelberger:
Please add some integration tests for this mode.

plugins/modules/openssh_cert.py
```diff
@@ -53,6 +53,11 @@
             - If this is set, I(signing_key) needs to point to a file containing the public key of the CA.
         type: str
         version_added: 1.1.0
+    use_agent:
+        description:
+            - Should the ssh-keygen use a CA key residing in a ssh-agent.
+        type: bool
+        default: false
```
felixfontein, line 60:
```suggestion
        default: false
        version_added: 1.2.0
```

felixfontein (commented):
You also need to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

dougstanley:
Thanks for the feedback. I'll add the tests and changelog fragment and update the pull request.

dougstanley:
So, it would seem that signing a cert using a CA stored in the ssh-agent was a feature added in openssh 7.6 and so isn't available on some older linux distros (it seems ubuntu 16.04 and centos 7 and older are the main ones).
Does that make this feature addition a no-go, or is there a way to specify that the feature is only available on certain hosts? I could make the parameter ignored or cause an error immediately on hosts that don't support it. If there are community standards for dealing with these situations, I apologize, this is my first attempt at a PR for anything ansible.

MarkusTeufelberger:
So far we tried to support the lowest common denominator (otherwise we would have dropped pyOpenSSL a long time ago, or support for running these modules under Python 2.6). RHEL 7 (CentOS keeps roughly the same timeline) will be supported until end of June 2024 according to https://access.redhat.com/support/policy/updates/errata/

There likely is a way to check for this feature and then act accordingly, but ideally it would be something that can be bacported in some other way and used on Ubuntu Xenial or CentOS 7 as well, so there's not too much feature disparity.

OTOH some features (e.g. handling certain key types, such as ed25519 ones) are only available using the `cryptography` backend which is not available on some distros.

More problematic might be that at least according to https://linux.die.net/man/1/ssh-keygen  the `-U` switch that you use is already in use for uploading keys into a smartcard reader.

As a way forward I guess you can think of a way how you would expect this to behave on e.g. Ubuntu Xenial and what the impact would be in a typical use case (e.g. should the signing key even ever be transferred onto a remote machine?). You could for example argue that the signing should mostly take place on the Ansible control machine, not some host out there and it might be easier for users to upgrade their workstation rather than all their servers.

felixfontein:
I think the most important part is that the user gets a helpful error message if used with a too old ssh-keygen version, and that it doesn't accidentally do something different. That `-U` is used for *different* things with different ssh-keygen versions is worrysome.

I checked out the git history of openssh-portable. The use of `-U` for smartcards was removed in https://github.com/openssh/openssh-portable/commit/7ea845e48df6d34a333ebbe79380cba0938d02a5 (so it looks like that was done for version 5.4), and the use of `-U` for gpg-agent was added in https://github.com/openssh/openssh-portable/commit/a98339edbc1fc21342a390f345179a9c3031bef7 (and this was apparently done for version 7.6).

With that information, it should be possible to implement a version test. Unfortunately, ssh-keygen itself seems to be incapable of outputting its version, so `ssh -V` needs to be called.

dougstanley:
I have added tests and also put in a version check so that if some one attempts to use the use_agent parameter, it will return an error stating that the minimum ssh version is 7.6. I also put version checks on all of the tests so if the tests don't run where they're guaranteed to fail.

Please let me know if anything doesn't look right. I tried to make sure I kept all the changes similar to how other things were implemented.

Also, is there a standard way to test for a known failure? In the tests, I did one where I expected a failure and just used `ignore_errors` on it, and followed it with an assert to make sure that it failed as expected. Let me know if there's a better way to do that. I'm also wondering if I should have a test to make sure that it fails appropriately IF the ssh version is too old.

felixfontein:
> Also, is there a standard way to test for a known failure? In the tests, I did one where I expected a failure and just used ignore_errors on it, and followed it with an assert to make sure that it failed as expected. Let me know if there's a better way to do that. I'm also wondering if I should have a test to make sure that it fails appropriately IF the ssh version is too old.

When the module failed, you can check `result.msg` for the error message. See for example https://github.com/ansible-collections/community.crypto/blob/main/tests/integration/targets/openssh_keypair/tasks/main.yml#L139-L158

plugins/modules/openssh_cert.py
```diff
@@ -53,6 +53,12 @@
...
+        version_added: 1.2.0
```
felixfontein, line 61:
```suggestion
        version_added: 1.3.0
```

plugins/module_utils/crypto/support.py
```diff
@@ -223,6 +223,18 @@ def parse_name_field(input_dict):
     return result
 
 
+def parse_openssh_version(version_string):
```
felixfontein, line 226: Could you put that into a new file `plugins/module_utils/openssh.py`?

plugins/modules/openssh_cert.py
```diff
@@ -273,6 +283,22 @@ def generate(self, module):
             if self.pkcs11_provider:
                 args.extend(['-D', self.pkcs11_provider])
 
+            if self.use_agent:
+                proc = module.run_command([self.ssh, '-Vq', self.path])
```
felixfontein, line 287: I think it's better to do the version check in `main()`, and fail there, before any other logic is executed. Then the behavior will be a lot more consistent, and it will always fail if `use_agent` is specified and the version is too old, instead of only failing in case the certificate needs to be (re-)generated.

tests/integration/targets/openssh_cert/tasks/main.yml
```diff
@@ -355,6 +355,85 @@
...
+      public_key: '{{ output_dir }}/id_key.pub'
+      path: '{{ output_dir }}/id_cert_with_agent'
+      use_agent: yes
+      valid_from: always
+      valid_to: forever
+    register: rc_no_key_in_agent
+    ignore_errors: yes
+    environment:
+      SSH_AUTH_SOCK: "{{ openssh_agent_sock }}"
+    when: openssh_version is version("7.6",">=")
```
felixfontein, line 371: Instead of adding `when:` in every step, how about using `block:` with `when:`?

plugins/module_utils/crypto/support.py
dougstanley: Done.

plugins/modules/openssh_cert.py
dougstanley: I moved the check to main as early as I could.

tests/integration/targets/openssh_cert/tasks/main.yml
dougstanley: Sorry, I hadn't actually used or come across the block keyword before, didn't realize how useful it would have been. I reorganized the tests and put them in a separate block.

changelogs/fragments/117-openssh_cert-use-ssh-agent.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- openssh_cert - add module parameter ``use_agent`` to enable using signing keys stored in ssh-agent. (https://github.com/ansible-collections/community.crypto/issues/116)
```
felixfontein, line 2:
```suggestion
- openssh_cert - add module parameter ``use_agent`` to enable using signing keys stored in ssh-agent (https://github.com/ansible-collections/community.crypto/issues/116).
```

tests/integration/targets/openssh_cert/tasks/main.yml
```diff
@@ -411,3 +411,87 @@
...
+    check_mode: yes
+  - name: Remove keypair for agent tests
+    openssh_keypair:
+      path: '{{ output_dir }}/id_key'
+      state: absent
+  - name: Remove certificate
+    openssh_cert:
+      state: absent
+      path: '{{ output_dir }}/id_cert_with_agent'
+      #type: user
```
felixfontein, line 492: Can you remove this and the following lines?

felixfontein (commented):
Looks good! Just two details :)

tests/integration/targets/openssh_cert/tasks/main.yml
dougstanley: So remove the entire task that removes the id_cert_with_agent cert, or just remove the commented lines? I wasn't sure if we were supposed to have cleanup tasks to remove things we created or not.

felixfontein: Just the commented out lines. They are not needed.

(Cleanup is also not necessary, but since the other tests for this module do that as well...)

dougstanley: I believe I have this all done. I *did* also forget to make the one test for failure more explicit, so I updated that test as well and included making sure the correct error was in the output. Hopefully that change is ok. Let me know if it needs any adjustment.

felixfontein (approved):
LGTM

@MarkusTeufelberger what do you think?

MarkusTeufelberger:
Yeah, looks good to me. Thanks for implementing the feature @dougstanley!

felixfontein:
@dougstanley thanks a lot for implementing this, especially with tests :)
@MarkusTeufelberger thanks a lot for reviewing!

dougstanley:
Thanks @felixfontein and @MarkusTeufelberger for your help!

PR #119, opened 2020-10-08
Refactor openssl_privatekey module, move  add openssl_privatekey_pipe module

felixfontein:
##### SUMMARY
`openssl_privatekey_pipe` is providing almost the same functionality than `openssl_privatekey`, except that it does not operate on disk. An existing private key can (optionally) be passed into the module (for idempotency checks, or for conversion), and the new/converted/unchanged private key is returned, but not written to disk.

This allows to store key data in vaults without having to write the unprotected key to disk (for example with the [community.sops collection](https://github.com/ansible-collections/community.sops)).

For this, I moved the main code from `openssl_privatekey` to plugins/module_utils/crypto/module_backends/privatekey.py, and the corresponding documentation to plugins/doc_fragments/module_privatekey.py.

I'm not 100% sure about the `_pipe` suffix. If someone has a better idea, feel free to mention it :)

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
openssl_privatekey
openssl_privatekey_pipe


MarkusTeufelberger:
Isn't this functionality (ephemeral state) something that's more likely part of an action plugin? Also visible in logs or not, this will always transmit the private key material to the Ansible controller and the target host. It seems to me that this is something that you shouldn't run on remote hosts anyways (you only might want to assign the task to the host to get some context/host_vars).

Implementing this as action plugin however is a LOT more pain than one might initially expect, just as a fair warning.

I don't have a better name than the `_pipe` suffix (maybe `_ephemeral`? Harder to spell correctly though and not much more correct...).

Maybe there's also the option of adding a `content` field to `openssl_privatekey` and the option to only return the key in the module output instead to get a similar functionality and stay in line with `openssl_privatekey_info`?

felixfontein:
@MarkusTeufelberger I'd think the module will be mostly used for `localhost` (i.e. `delegate_to: localhost`); maybe I should add that to the example to make this clearer.

Making it an action plugin is a good idea in this case, though I wanted to do the same extension to other modules as well, most prominently `openssl_csr`, and for that an action plugin would be very limiting, since the private key input both makes sense as content and a file. You could execute `openssl_csr_pipe` on a remote host (which has the private key stored on disk), without having to write the `.csr` to disk.

About adding a `content` field to `openssl_privatekey`: that's possible, but very painful to implement, since the module is built around the idea of operating on a file on disk. I think having another module (or action plugin) specialized on this use-case is better.

> I don't have a better name than the `_pipe` suffix (maybe `_ephemeral`? Harder to spell correctly though and not much more correct...).

I agree that `_ephemeral` is harder to spell correctly, so let's avoid that. On the other hand, it would be a good training exercise for me, so I will hopefully misspell it less often in the future ;)

felixfontein:
I'll look into changing this to an action plugin later. I think it should not be that hard, with a proper bit of framework code for argument validation. Fortunately most of that has already been moved out ot AnsibleModule into generic validation functionality in https://github.com/ansible/ansible/tree/devel/lib/ansible/module_utils/common/

felixfontein:
I've changed the module to an action plugin. For that, I've added a small framework (mostly copy'n'pasta from ansible.module_utils.basic) which provides similar argument validation facilities as AnsibleModule.

felixfontein:
ready_for_review

plugins/module_utils/action_module.py
```diff
@@ -0,0 +1,733 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2012-2013 Michael DeHaan <michael.dehaan@gmail.com>
+# Copyright (c) 2016 Toshio Kuratomi <tkuratomi@ansible.com>
+# Copyright (c) 2019 Ansible Project
+# Copyright (c) 2020 Felix Fontein <felix@fontein.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+# Parts taken from ansible.module_utils.basic and ansible.module_utils.common.warnings.
```
MarkusTeufelberger, line 9: As I understand it, the code in this file is part of Ansible 2.11 and should be dropped from this collection once we don't support Ansible 2.10 any more, right?

plugins/action/openssl_privatekey_pipe.py
```diff
@@ -0,0 +1,110 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+import base64
+
+from ansible.module_utils._text import to_native, to_bytes
+
+from ansible_collections.community.crypto.plugins.module_utils.action_module import ActionModuleBase
```
MarkusTeufelberger, line 14: If I understand it correctly, this helper stuff is part of Ansible 2.11 already - maybe we could make this import conditional to Ansible <=2.10?

plugins/module_utils/crypto/module_backends/privatekey.py
```diff
@@ -0,0 +1,590 @@
...
+            'create': create,
+            'verify': verify,
+            'deprecated': deprecated,
+        }
+
+    def __init__(self, module):
+        super(PrivateKeyCryptographyBackend, self).__init__(module=module, backend='cryptography')
+
+        self.curves = dict()
+        self._add_curve('secp384r1', 'SECP384R1')
```
MarkusTeufelberger, line 315: The ordering of these curves triggers my inner Adrian Monk... :-D

plugins/module_utils/crypto/module_backends/privatekey.py
```diff
@@ -0,0 +1,590 @@
...
+                             exception=CRYPTOGRAPHY_IMP_ERR)
+        return backend, PrivateKeyCryptographyBackend(module)
+    else:
+        raise Exception('Unsupported value for backend: {0}'.format(backend))
+
+
+def get_privatekey_argument_spec():
+    return (
+        # argument_spec
+        dict(
```
MarkusTeufelberger, line 560: Why do we use `dict(foo='bar')` that often in module argument code by the way? Just curious, since I don't see how `{'foo': 'bar'}` would be different and this style seems to be mostly used in module arguments in Ansible modules.

plugins/modules/openssl_privatekey.py
```diff
@@ -275,610 +140,115 @@
...
+        argument_spec=argument_spec,
         supports_check_mode=True,
         add_file_common_args=True,
-        required_together=[
-            ['cipher', 'passphrase']
-        ],
-        required_if=[
-            ['type', 'ECC', ['curve']],
-        ],
+        required_together=required_together + [],
```
MarkusTeufelberger, line 250: Not sure if it is necessary to explicitly mention this extensibility.

plugins/modules/openssl_privatekey_pipe.py
```diff
@@ -0,0 +1,112 @@
...
+  community.crypto.openssl_privatekey_pipe:
+    content: "{{ lookup('community.sops.sops', 'private_key.pem.sops') }}"
+    size: 2048
+  register: output
+  no_log: true  # make sure that private key data is not accidentally revealed in logs!
+- name: Update encrypted key when openssl_privatekey_pipe reported a change
+  community.sops.encrypt_sops:
+    path: private_key.pem.sops
+    content_text: output.privatekey
+  when: output is changed
```
MarkusTeufelberger, line 72: Ansible-lint would mark this as "This should be a handler". ;-)

I'm not sure what the impact running this every time instead of only on changes would be, maybe this `when` can just be removed?

MarkusTeufelberger (commented):
I didn't find much, maybe a few ideas to consider or things to clarify. Hope it helps, I think it can also be merged as-is.

plugins/module_utils/action_module.py
felixfontein: Not exactly. The code is already in Ansible 2.9, but unfortunately it is part of `AnsibleModule` and cannot be easily reused without hacks. That's why I copy it here. There are plans to extract more of the argument spec validation; once that happens, I hope we can reduce much of the code from this file. Until then, we probably have to live with copies if we don't want to resort to hacks.

plugins/action/openssl_privatekey_pipe.py
felixfontein: ActionModuleBase does not exist like this in any ansible-base version to my knowledge.

plugins/module_utils/crypto/module_backends/privatekey.py
felixfontein: Good question. I've seen both. In the modules I happened to work on more intensively, in almost all (or even all?) cases the `dict()` syntax was used.

(Also, in this case, this is just copy'n'paste from openssl_privatekey.)

plugins/modules/openssl_privatekey_pipe.py
felixfontein: > Ansible-lint would mark this as "This should be a handler". ;-)

If that's true, this is one of the cases very ansible-lint is very wrong. Replacing such code by a handler is a really stupid idea. There are a lot of situations where handlers make sense, but also a lot where they don't. This is definitely one of the latter.

> I'm not sure what the impact running this every time instead of only on changes would be, maybe this `when` can just be removed?

It depends on whether you have set `return_current_key: true` in the `openssl_privatekey_pipe` call. If you didn't, `output.privatekey` would be `VALUE_SPECIFIED_IN_NO_LOG_PARAMETER` is `output is not changed`, and you really don't want to write that into your private key file.

plugins/action/openssl_privatekey_pipe.py
MarkusTeufelberger: :crying_cat_face: 

Let's hope something like it eventually ends up upstream - action plugins are unnecessarily hard to write in some cases.

plugins/modules/openssl_privatekey_pipe.py
MarkusTeufelberger: https://ansible-lint.readthedocs.io/en/latest/default_rules.html#tasks-that-run-when-changed-should-likely-be-handlers

I am also a bit torn about this linting rule, just something that I came across quite often and a pattern that I recognize now when I see it.

plugins/module_utils/crypto/module_backends/privatekey.py
felixfontein: The order is equivalent to here: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#elliptic-curves except that originally I forgot to add SECP256R1, and had to add it later - and then apparently added it next to SECP256K1.

Which order do you prefer? Lexicographic? Bitsize, then lexicographic? Recommended first? ...?

plugins/modules/openssl_privatekey_pipe.py
felixfontein: I'd say that in maybe half of the use-cases of `when: x is changed` it is totally correct. But in the other half, it is just wrong. You can partially rescue things with `meta: flush_handlers`, but that can make things even messier.

plugins/action/openssl_privatekey_pipe.py
felixfontein: I hope so too. I know that more of the argument spec handling will get extracted for ansible-base 2.11, since there will be an action plugin for validating role argument specs. It will be interesting to see how that looks in the end, and how does it help with this task.

plugins/modules/openssl_privatekey.py
felixfontein: I've removed it.

plugins/module_utils/crypto/module_backends/privatekey.py
felixfontein: Maybe let's change all of them in this collection for community.crypto 2.0.0 :)

plugins/module_utils/crypto/module_backends/privatekey.py
MarkusTeufelberger: I would sort first by not deprecated/deprecated, then by curve type (Koblitz, Brainpool etc.) then by bitsize (smallest first).

https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec.html#elliptic-curves also states that curves should have at least 224 bits, so probably we should also "deprecate" secp192r1.

MarkusTeufelberger: Thinking more about it (there's also a list of algorithms in the module's documentation) this probably is better done separately.

MarkusTeufelberger:
Looks good to me, thanks for all the work!

/shipit

plugins/module_utils/crypto/module_backends/privatekey.py
felixfontein: Ok, I'll create a follow-up PR for that :)

felixfontein:
@MarkusTeufelberger thanks a lot for reviewing this!

PR #132, opened 2020-10-28
ECC curve list order

felixfontein:
##### SUMMARY
As @MarkusTeufelberger requested in #119.

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
plugins/doc_fragments/module_privatekey.py
plugins/module_utils/crypto/module_backends/privatekey.py


MarkusTeufelberger (approved):
"K" and "R" curves are not as similar as this sorting might make it look like, but I guess it is still easy enough to see if a desired curve is (not) supported.

Thanks for the fix!

felixfontein:
There's also always Ctrl+F ;-)

felixfontein:
@MarkusTeufelberger thanks for reviewing this!

PR #134, opened 2020-11-01
Move action_module from module_utils to plugin_utils

felixfontein:
##### SUMMARY
`module_utils` is meant for things that can be included by both modules and plugins. This is why I had to suppress several sanity errors when putting `action_module.py` there (in #119), which is only meant to be used by plugins.

This PR moves that file to a new directory `plugins/plugin_utils`. Ansible doesn't know about this directory, but I guess there's no need, since Ansible only has to know about places were to grab files for modules from (to send them to the remote system), and not for files used by plugins.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/action_module.py


felixfontein:
I asked a bit on #ansible-devel on whether this is a good idea. The summary is "module_utils is good in general, but for plugin-only code everything else can be used (resp. the collection loader allows everything), but it would be good to have an official place for it". Anyway, I created a proposal (ansible/proposals#186) to suggest plugins/plugin_utils.

felixfontein:
I'll just merge it since it works. If another official place than plugins/plugin_utils is designated eventually, we can move again.

PR #135, opened 2020-11-01
Refactor x509_certificate module, add x509_certificate_pipe module

felixfontein:
##### SUMMARY
Similar refactoring / addition than in #119 and #123: refactor x509_certificate module to move large parts of documentation to doc fragments and large pieces of code to module backends, in this case split up per provider, and add a new x509_certificate_pipe module which does not operate on certificate files (and supports only a subset of the providers, maybe just selfsigned and ownca in the beginning).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
x509_certificate
x509_certificate_pipe


felixfontein:
ready_for_review

@ctrufan could you test whether with this PR the `entrust` provider still works for `x509_certificate`?

ctrufan:
Probably not until later this week or this weekend unfortunately, need to rebuild my testing VM.

felixfontein:
Tests should run better now that #137 is done and merged. The remaining failed run is because RHEL8 is currently broken in CI (unrelated to any PR).

felixfontein:
@ctrufan did you got around to test this already?

felixfontein:
(I rebased to be able to work with this module with the bugfix from #141.)

ctrufan:
Sorry, ran into some issues getting my environment set up to test with 2.10 and the new collections paradigm.

Do I need to run the tests from the primary ansible subdirectory? I tried running from the checked out branch and got an error. The documentation doesn't quite make it clear how to run tests within a collection-from-code.

{code}
The current working directory must be at or below:

-  The ansible source: /ansible/ansible
- An Ansible collection {...}/ansible_collections/${namespace}/${collection}
{code}

ctrufan:
> @felixfontein
> Sorry, ran into some issues getting my environment set up to test with 2.10 and the new collections paradigm.
> 
> Do I need to run the tests from the primary ansible subdirectory? I tried running from the checked out branch for openssl_certificate_pipe and got an error. The documentation doesn't quite make it clear how to run tests within a collection-from-code.
> 
> ```
> The current working directory must be at or below:
> 
> -  The ansible source: /ansible/ansible
> - An Ansible collection {...}/ansible_collections/${namespace}/${collection}
> ```

Nevermind, I didn't have quite the right folder structure. Working on testing now/today. Sorry about the unnecessary ping!

ctrufan:
It's not working even pre-pull request because entrust_certificate was broken when the additions were made to support providing the full cert instead of path for 2.10/community migration.

```
 if self.csr_content is None **or** not os.path.exists(self.csr_path):
            raise CertificateError(
                'The certificate signing request file {0} does not exist'.format(self.csr_path)
            )
```
vs, for example, ownca:
```
if self.csr_content is None **and** not os.path.exists(self.csr_path):
            raise CertificateError(
                'The certificate signing request file {0} does not exist'.format(self.csr_path)
            )
```

plugins/module_utils/crypto/module_backends/certificate_entrust.py
```diff
@@ -0,0 +1,221 @@
...
+    def __init__(self, module, backend):
+        super(EntrustCertificateBackend, self).__init__(module, backend)
+        self.trackingId = None
+        self.notAfter = get_relative_time_option(module.params['entrust_not_after'], 'entrust_not_after', backend=self.backend)
+
+        if self.csr_content is None and self.csr_path is None:
+            raise CertificateError(
+                'csr_path or csr_content is required for entrust provider'
+            )
+        if self.csr_content is None or not os.path.exists(self.csr_path):
```
ctrufan, line 50:
```suggestion
        if self.csr_content is None and not os.path.exists(self.csr_path):
```

felixfontein:
Hmm, that bad condition comes from ansible/ansible#66384. Looks like we totally overlooked it back then...

felixfontein:
I guess we should sooner or later add some kind of tests for the ECS modules. Integration tests will need some kind of API simulator. Alternatively, this could be done as a unit test which patches the `Request.open()` method (similar to the framework in https://github.com/ansible-collections/community.internal_test_tools/blob/main/tests/unit/utils/fetch_url_module_framework.py; see https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/net_tools/test_hetzner_firewall.py#L69-L111 for a concrete test case of a module with that framework).

ctrufan:
I like the idea of returning a mocked response, but it would get a bit tricky, because there's a call to download the API spec unless it's provided locally (that's unauthenticated, but I don't think it would be great to have a unit test download a large file hah).

So the easier route to go might be to mock the module_util return somehow.

We do have integration tests on 'our' side, but I neglected to update them for the new collections paradigm, which requires quite a few tweaks to the molecule framework.

felixfontein:
I've run the new acme provider tests from #142 with this PR and they passed. So at least the acme provider still seems to work with this PR ;)

felixfontein:
@ctrufan with large file, do you mean https://cloud.entrust.net/EntrustCloud/documentation/cms-api-2.1.0.yaml ? It's "only" 110 kb, so downloading it during the tests should be ok. (Assuming your server doesn't mind.) It's probably best to download it once during test setup, and then keep the file locally.

In any case, mocking the module_utils is definitely easier. And it still allows us to do some modifications to the caller and being half-way confident that it still works ;)

ctrufan:
Previously, if the certificate file pointed to by path was an empty or invalid file, the module still worked, and replaced it with a valid certificate. That's no longer the case, but I think that's valid behavior.

Running into some issues getting the test CA recognized properly after updating Python, because using custom CA certs is an OS/python version dependent nightmare, so might not have the valid test case through until tomorrow - module is working correctly up through the cert request stage though.

ctrufan:
Yeah our server has no problem with multiple downloads of the spec.

Something that's been on my to-do list for ages is to update the module so that it can 'cache' the API spec, but it's not something (last I looked) that Ansible has elegant solutions for.

felixfontein:
> Previously, if the certificate file pointed to by path was an empty or invalid file, the module still worked, and replaced it with a valid certificate. That's no longer the case, but I think that's valid behavior, it's only "not backwards compatible" in a "this probably shouldn't have worked anyways" sense.

That wasn't an intended change - I've restored the old behavior in dce4f76e55dd06b6b481721eab701d24de622df5.

> Running into some issues getting the test CA recognized properly after updating Python, because using custom CA certs is an OS/python version dependent nightmare,

Indeed!

> so might not have the valid test case through until tomorrow - module is working correctly up through the cert request stage though.

BTW, I recently learned of https://trustme.readthedocs.io/en/latest/ (ansible-community/antsibull#133). Haven't gotten around to really look into it yet, but it might help with such things. Or at least its internals might be useful :)

felixfontein:
> Something that's been on my to-do list for ages is to update the module so that it can 'cache' the API spec, but it's not something (last I looked) that Ansible has elegant solutions for.

Since modules can run anywhere, it's kind of hard to cache something reliably. (And most users don't want modules to create files in random positions they didn't explicitly ask for.) Maybe there could be a module `entrust_cache_apispec` to handle only that? Not sure how practical / useful that is though...

ctrufan:
> BTW, I recently learned of https://trustme.readthedocs.io/en/latest/ ([ansible-community/antsibull#133](https://github.com/ansible-community/antsibull/issues/133)). Haven't gotten around to really look into it yet, but it might help with such things. Or at least its internals might be useful :)

That's a neat project, I'll have to look at it more indepth sometime, but I suspect it solves a different problem - I've got a robust test CA infrastructure set up, CRLs/OCSP included, but the underlying HTTP calls made by the python url libraries like to insist there's no such thing as an operating system or editable trust store, especially if certifi is installed.

ctrufan:
> ince modules can run anywhere, it's kind of hard to cache something reliably. (And most users don't want modules to create files in random positions they didn't explicitly ask for.) Maybe there could be a module `entrust_cache_apispec` to handle only that? Not sure how practical / useful that is though...

That was my thought process when I was first developing the module and wanted to cache - don't want to create a file the user isn't expecting, can't easily store it in memory because of how modules run anywhere.

ctrufan:
@felixfontein 

Tested the module, and it looks good, worked with entrust provider after the fixes. (I'm assuming from PR description entrust doesn't leverage the new "pipe" module, so there's no new input/output paradigms I need to test?)

Though now I definitely need to carve time out to add unit tests and update our integration tests to work properly with collections, newer python min version, etc... and probably update x509 w/ entrust provider, and ecs_certificate to take advantage of similar new functionality...

felixfontein:
@ctrufan thanks for testing!!! I noticed that I wrote the PR description some time ago when I didn't actually implement the _pipe module; right now the _pipe module also supports the entrust provider, but from the way the unified provider interface works, it should probably just work. But then, it's still better to check it I guess ;) I guess there are two choices: a) remove the entrust provider - that's a very simple change. b) you actually test it.

I'd like to get this merged soon, so I guess if you don't have time now to test it, it's better to remove it for now. Re-adding it later is very easy, so we can do that for the next release without much work involved. (Well, at least not for me - it's mostly work for you, thanks to testing :) ) What do you think?

plugins/modules/x509_certificate_pipe.py
```diff
@@ -0,0 +1,171 @@
...
+    content: "{{ lookup('file', '/etc/ssl/csr/www.ansible.com.crt') }}"
+    csr_content: "{{ lookup('file', '/etc/ssl/csr/www.ansible.com.csr') }}"
+    ownca_cert: /path/to/ca_cert.crt
+    ownca_privatekey: /path/to/ca_cert.key
+    ownca_privatekey_passphrase: hunter2
+  register: result
+- name: Store certificate
+  ansible.builtin.copy:
+    path: /etc/ssl/csr/www.ansible.com.crt
+    content: "{{ result.certifiate }}"
```
ctrufan, line 75:
```suggestion
    content: "{{ result.certificate }}"
```

ctrufan:
It looks like if I use x509_certificate_pipe, requests fail when I pass file contents in with lookup(), because the value of "csr" that gets populated in the API request that goes out to the server is bytes rather than a string.

It works if I throw an arbitrary to_text in the line of entrust_certificate around line 100 (along with an import of to_text at the top)
`body['csr'] = to_text(self.csr_content)`
But that was just a sort of makeshift "see if this fixes it" approach.

I think a better answer might be that I should be converting all the values in the array that gets passed to the ECS api.py file to normalize them to text. Either that or doing something on first parameter load.

I don't think I'll have time to investigate any farther until tomorrow, so feel free to merge it with the pipe removed for now, I don't want to delay your merge.

felixfontein:
> It looks like if I use x509_certificate_pipe, requests fail when I pass file contents in with lookup(), because the value of "csr" that gets populated in the API request that goes out to the server is bytes rather than a string.

Actually x509_certificate would fail the same way if you use `csr_content`. So it's something to fix.

> It works if I throw an arbitrary to_text in the line of entrust_certificate around line 100 (along with an import of to_text at the top)
> `body['csr'] = to_text(self.csr_content)`
> But that was just a sort of makeshift "see if this fixes it" approach.

This is the correct fix (except that using `to_native` is even better), since `csr_content` is converted to `bytes` because that's the format it is usually needed in (for loading the CSR, or writing it to disk for the `acme` provider).

> I think a better answer might be that I should be converting all the values in the array that gets passed to the ECS api.py file to normalize them to text. Either that or doing something on first parameter load. I'll have to revisit the ansible guidelines on this stuff and Python 2/3, all the string/byte stuff is not something I recall without looking it up again!

For everything that is taken from `module.params` that should not be necessary, they should already be in the correct format. It's only that `csr_content` is encoded in the `CertificateBackend` base class since the providers (and the base class) usually need it as `bytes`. Well, except the `entrust` provider in this one case :)

> I don't think I'll have time to investigate any farther until tomorrow, so feel free to merge it with the provider removed for now, I don't want to delay your merge.

I'll still wait a few days, in the hope that @MarkusTeufelberger also finds some time to look at this :)

plugins/modules/x509_certificate_pipe.py
```diff
@@ -0,0 +1,161 @@
...
+    provider: ownca
+    content: "{{ lookup('file', '/etc/ssl/csr/www.ansible.com.crt') }}"
+    csr_content: "{{ lookup('file', '/etc/ssl/csr/www.ansible.com.csr') }}"
+    ownca_cert: /path/to/ca_cert.crt
+    ownca_privatekey: /path/to/ca_cert.key
+    ownca_privatekey_passphrase: hunter2
+  register: result
+- name: Store certificate
+  ansible.builtin.copy:
+    path: /etc/ssl/csr/www.ansible.com.crt
```
ctrufan, line 71:
```suggestion
    dest: /etc/ssl/csr/www.ansible.com.crt
```

Path throws errors, at least in my current version of ansible 2.10

felixfontein: Thanks for spotting that! There's the same error in openssl_csr_pipe as well...

ctrufan:
>It's only that csr_content is encoded in the CertificateBackend base class since the providers (and the base class) usually need it as bytes. Well, except the entrust provider in this one case :)

Ah, I see, I missed that it was explicitly being converted, thought I had just messed up normalizing somewhere. Yeah, loading the contents bytes are appropriate for most of the module use cases, we just happen to need to provide it as an API parameter.

I confirmed that having to_native fixes both x509_Certificate and x509_Certificate_pipe, I was able to reference both cert contents and csr contents as a stream with your latest update (well, after hacking to re-add entrust provider).

felixfontein:
@ctrufan thanks a lot for testing this! If you want, I can re-add the entrust provider to x509_certificate_pipe. (We can also wait for a later version if you prefer that.)

ctrufan:
No problem, sorry it took so long to get everything tested!

Re-adding would be great, thanks! (Not a big deal if you don't get around to it).

felixfontein:
@ctrufan thanks for testing! This here is a voluntary community project, so no need to apoligize!

Re-adding is easy, I just had to revert that (simple) commit ;-)

felixfontein:
I rebased to `main` and added an empty `ci_complete` commit to make sure that all tests are run.

felixfontein:
@xyon @Shaps @MarkusTeufelberger @puiterwijk would be happy if one of you could take a look at this PR!

MarkusTeufelberger:
Sorry for the radio silence, bit busy times right now... :-/

I'll take a look at it in the evening today.

plugins/modules/x509_certificate_pipe.py
```diff
@@ -0,0 +1,171 @@
...
+
+DOCUMENTATION = r'''
+---
+module: x509_certificate_pipe
+short_description: Generate and/or check OpenSSL certificates
+version_added: 1.3.0
+description:
+    - It implements a notion of provider (ie. C(selfsigned), C(ownca), C(entrust))
+      for your certificate.
+    - "Please note that the module regenerates existing certificate if it doesn't match the module's
```
MarkusTeufelberger, line 21:
```suggestion
    - "Please note that the module regenerates an existing certificate if it doesn't match the module's
```

MarkusTeufelberger (commented):
I mostly looked at documentation and kinda hope that immediate failures would be caught by the integration tests. Hope it still helps, thanks for the great work as always!

plugins/modules/x509_certificate_pipe.py
```diff
@@ -0,0 +1,171 @@
...
+- name: Generate a Self Signed OpenSSL certificate
+  community.crypto.x509_certificate_pipe:
+    provider: selfsigned
+    privatekey_path: /etc/ssl/private/ansible.com.pem
+    csr_path: /etc/ssl/csr/ansible.com.csr
+  register: result
+- debug:
+    var: result.certificate
+
+- name: Generate an OpenSSL Certificate Signing Request with an inline certificate
```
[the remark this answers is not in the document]
felixfontein: Heh, it should have been `Generate certificate with inline CSR` :) I've fixed that. I also added more comments and a `delegate_to` to the `copy` task. This example should show that while the OwnCA data (private key, certificate) is on the remote machine, we don't have to copy CSR and certificate to/from there.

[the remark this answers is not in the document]
felixfontein: I reformatted the example and added another one as you suggested.

plugins/doc_fragments/module_certificate.py
```diff
@@ -0,0 +1,575 @@
...
+            - Note that this is only supported if the C(cryptography) backend is used!
+        type: bool
+        default: yes
+'''
+
+    BACKEND_SELFSIGNED_DOCUMENTATION = r'''
+options:
+    csr_path:
+        description:
+            - This is not required for the C(selfsigned) provider.
```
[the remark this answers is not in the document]
felixfontein: I thought a bit about this and just made it more verbose. I think it's better to mention that not specifying it is usually not what you want.

Since this fragment is only used together with the main fragment (in `DOCUMENTATION`), the part about being mutually exclusive is already there. This only adds the `selfsigned`-specific information.

felixfontein:
recheck

felixfontein:
@MarkusTeufelberger if you're happy with the changes, I'll merge (and release 1.3.0 I guess, if nobody objects).

plugins/doc_fragments/module_certificate.py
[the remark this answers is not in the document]
felixfontein: I noticed that this part is overwritten by the 'generic' text (which contains the part about mutually exclusive), so I removed this one completely.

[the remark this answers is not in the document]
felixfontein: (And converted it to a `notes:` entry.)

plugins/module_utils/crypto/module_backends/certificate.py
```diff
@@ -0,0 +1,392 @@
...
+        if not PYOPENSSL_FOUND:
+            module.fail_json(msg=missing_required_lib('pyOpenSSL >= {0}'.format(MINIMAL_PYOPENSSL_VERSION)),
+                             exception=PYOPENSSL_IMP_ERR)
+        if provider.needs_pyopenssl_get_extensions(module):
+            try:
+                getattr(crypto.X509Req, 'get_extensions')
+            except AttributeError:
+                module.fail_json(msg='You need to have PyOpenSSL>=0.15')
+
+        module.deprecate('The module is using the PyOpenSSL backend. This backend has been deprecated',
```
resmo, line 360: not a blocker but wouldn't it make sense to move this before right after line 350 (before conditions) so the deprecation warning should be shown even if the lib is missing?

plugins/module_utils/crypto/module_backends/certificate.py
```diff
@@ -0,0 +1,392 @@
...
+            module.fail_json(msg='The cryptography backend does not support v2 certificates, '
+                                 'use select_crypto_backend=pyopenssl for v2 certificates')
+
+    return provider.create_backend(module, backend)
+
+
+def get_certificate_argument_spec():
+    return ArgumentSpec(
+        argument_spec=dict(
+            provider=dict(type='str', choices=[]),
```
resmo, line 377: ehm is this as it should?

plugins/module_utils/crypto/module_backends/certificate.py
```diff
@@ -0,0 +1,392 @@
...
+            force=dict(type='bool', default=False,),
+            csr_path=dict(type='path'),
+            csr_content=dict(type='str'),
+            select_crypto_backend=dict(type='str', default='auto', choices=['auto', 'cryptography', 'pyopenssl']),
+
+            # General properties of a certificate
+            privatekey_path=dict(type='path'),
+            privatekey_content=dict(type='str', no_log=True),
+            privatekey_passphrase=dict(type='str', no_log=True),
```
resmo, line 386: :ok_hand: no_log=True  ;)

plugins/module_utils/crypto/module_backends/certificate.py
felixfontein: Yes: the `choices` list is updated for example here: https://github.com/ansible-collections/community.crypto/pull/135/files#diff-ccf70653d5d26e8f8b623afd6fccc608043ef7ac3b6481e4a9851a0c74338409R113

felixfontein: (Maybe a comment would help, though :) )

plugins/module_utils/crypto/module_backends/certificate.py
felixfontein: Fixed in 1b74d24a7c685129e8e4b7160232c2ac3f4c3763.

plugins/module_utils/crypto/module_backends/certificate.py
felixfontein: Added 14a0ca79f6d98fd98f7c7ddc6aa159d3896da6ec

resmo (approved):

felixfontein:
@ctrufan @MarkusTeufelberger @resmo thanks a lot for all the testing and reviewing!!!

PR #142, opened 2020-11-10
Add x509_certificate tests for ACME provider

felixfontein:
##### SUMMARY
Adds some tests for x509_certificate's ACME provider using a monkey-patched acme-tiny and the acme-test-container also used for the acme_* modules.

Note that I used another test target because right now, `cloud/acme` is not compatible with `setup_openssl` (needs some changes to ansible-test itself). This means that these tests will not be run automatically on a PR modifying x509_certificate, but only when all tests are run - for example during nightly test runs.

So this is not perfect, but still better than before :)

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
x509_certificate

### DIFF
--- a/tests/integration/targets/x509_certificate-acme/tasks/impl.yml
+++ b/tests/integration/targets/x509_certificate-acme/tasks/impl.yml
@@ -34,6 +34,17 @@
   environment:
     PATH: '{{ lookup("env", "PATH") }}:{{ output_dir }}'
 
+- name: Get certificate information
+  x509_certificate_info:
+    path: '{{ output_dir }}/cert-1.pem'
+  register: result
+
+- name: Validate certificate information
+  assert:
+    that:
+      - result.subject_alt_name | length == 1
+      - "'DNS:example.com' in result.subject_alt_name"
+
 - name: Retrieve certificate 2
   x509_certificate:
     provider: acme
@@ -44,3 +55,15 @@
     acme_directory: https://{{ acme_host }}:14000/dir
   environment:
     PATH: '{{ lookup("env", "PATH") }}:{{ output_dir }}'
+
+- name: Get certificate information
+  x509_certificate_info:
+    path: '{{ output_dir }}/cert-1.pem'
+  register: result
+
+- name: Validate certificate information
+  assert:
+    that:
+      - result.subject_alt_name | length == 2
+      - "'DNS:example.com' in result.subject_alt_name"
+      - "'DNS:example.org' in result.subject_alt_name"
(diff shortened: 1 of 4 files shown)


felixfontein:
Note that with all the monkey-patching of acme-tiny necessary, this will break sooner or later. I think it's still worth the trouble, to avoid random breakage of the acme provider in the future.

We could also eventually replace the acme-tiny usage by the ACME code from the acme_* modules. Then we wouldn't need acme-tiny anymore.

tests/integration/targets/x509_certificate-acme/tasks/impl.yml
```diff
@@ -0,0 +1,46 @@
+---
+- name: "Generate account key"
```
Andersson007, line 2:
```suggestion
- name: Generate account key
```
I'm not against quotes .. where necessary:D
later they are not used. One approach would look (imo) better:)

tests/integration/targets/x509_certificate-acme/tasks/main.yml
```diff
@@ -0,0 +1,123 @@
...
+      - intermediate
+    root_numbers:
+      - 0
+    interesting_keys:
+      - authority_key_identifier
+      - subject_key_identifier
+      - issuer
+      - subject
+      #- serial_number
+      #- public_key_fingerprints
```
Andersson007, line 56:
```suggestion
      #- serial_number
      #- public_key_fingerprints
```
is it like "TODO" or we can remove this?

Andersson007 (commented):
Any tests look good to me;)
I don't see any assertions or they will just fail if not ok?

tests/integration/targets/x509_certificate-acme/tasks/main.yml
felixfontein: Fixed in 4c70b27b00a60f72ad246c05e82edc9d9ddbcfd7.

tests/integration/targets/x509_certificate-acme/tasks/impl.yml
felixfontein: Fixed in 4c70b27b00a60f72ad246c05e82edc9d9ddbcfd7.

felixfontein:
They will simply fail if it doesn't work. Nonetheless I added 4ceffc284a62ab798b3e42bb097cce015f3bb6b0 to also do some checks on the results. Better safe than sorry :)

Andersson007 (approved):
Great, thanks for this!
LGTM

felixfontein:
@Andersson007 thanks a lot for reviewing this!

PR #172, opened 2021-01-16
Added sever name option to use for SNI

NorthFuture:
##### SUMMARY
when connecting to a server with SNI enabled and serving multiple certificates over the same hostname or ip, current implementation returns the wrong certificate, since server name is set to the hostname. This PR provides an additional option server_name that can be used to set the server name

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
get_certificate.py

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/get_certificate.py
+++ b/plugins/modules/get_certificate.py
@@ -306,7 +306,7 @@ def main():
             if not server_name:
                 server_name = host
                          
-            cert = ctx.wrap_socket(sock, server_hostname=server_name).getpeercert(True)
+            cert = ctx.wrap_socket(sock, server_hostname=server_name or host).getpeercert(True)
             cert = DER_cert_to_PEM_cert(cert)
         except Exception as e:
             if proxy_host:
--- a/plugins/modules/get_certificate.py
+++ b/plugins/modules/get_certificate.py
@@ -35,6 +35,10 @@ options:
         - The port to connect to
       type: int
       required: true
+    server_name:
+      description:
+       - Server name used for SNI when hostname is an ip or is different from server name
+      type: str
     proxy_host:
       description:
         - Proxy host used when get a certificate.
@@ -211,6 +215,7 @@ def main():
             port=dict(type='int', required=True),
             proxy_host=dict(type='str'),
             proxy_port=dict(type='int', default=8080),
+            server_name=dict(type='str'),
             timeout=dict(type='int', default=10),
             select_crypto_backend=dict(type='str', choices=['auto', 'pyopenssl', 'cryptography'], default='auto'),
         ),
@@ -222,6 +227,7 @@ def main():
     proxy_host = module.params.get('proxy_host')
     proxy_port = module.params.get('proxy_port')
     timeout = module.params.get('timeout')
+    server_name = module.params.get('server_name')
 
     backend = module.params.get('select_crypto_backend')
     if backend == 'auto':
@@ -297,7 +303,10 @@ def main():
                 ctx.check_hostname = False
                 ctx.verify_mode = CERT_NONE
 
-            cert = ctx.wrap_socket(sock, server_hostname=host).getpeercert(True)
+            if not server_name:
+                server_name = host
+                         
+            cert = ctx.wrap_socket(sock, server_hostname=server_name).getpeercert(True)
             cert = DER_cert_to_PEM_cert(cert)
         except Exception as e:
             if proxy_host:
--- a/plugins/modules/get_certificate.py
+++ b/plugins/modules/get_certificate.py
@@ -37,7 +37,7 @@ options:
       required: true
     server_name:
       description:
-       - Server name used for SNI when hostname is an ip or is different from server name
+       - Server name used for SNI (L(Server Name Indication,https://en.wikipedia.org/wiki/Server_Name_Indication)) when hostname is an IP or is different from server name.
       type: str
       version_added: 1.4.0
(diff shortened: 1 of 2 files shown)


NorthFuture:
What's the best strategy to test this?  @felixfontein

plugins/modules/get_certificate.py
```diff
@@ -35,6 +35,10 @@
         - The port to connect to
       type: int
       required: true
+    server_name:
+      description:
+       - Server name used for SNI when hostname is an ip or is different from server name
```
felixfontein, line 40:
```suggestion
       - Server name used for SNI (L(Server Name Indication,https://en.wikipedia.org/wiki/Server_Name_Indication)) when hostname is an IP or is different from server name.
```

plugins/modules/get_certificate.py
```diff
@@ -35,6 +35,10 @@
...
+      type: str
```
felixfontein, line 41:
```suggestion
      type: str
      version_added: 1.4.0
```

plugins/modules/get_certificate.py
```diff
@@ -297,7 +303,10 @@ def main():
                 ctx.check_hostname = False
                 ctx.verify_mode = CERT_NONE
 
-            cert = ctx.wrap_socket(sock, server_hostname=host).getpeercert(True)
+            if not server_name:
+                server_name = host
+                         
+            cert = ctx.wrap_socket(sock, server_hostname=server_name).getpeercert(True)
```
felixfontein, line 309: How about removing the `if` above, and using the following instead:
```suggestion
            cert = ctx.wrap_socket(sock, server_hostname=server_name or host).getpeercert(True)
```

plugins/modules/get_certificate.py
felixfontein: (We should probably add an explanation/link for SNI in the module description as well.)

felixfontein:
Please note the failing sanity tests:
```
plugins/modules/get_certificate.py:308:0: trailing-whitespace: Trailing whitespace
```

About testing: I'm not sure probably testing this is possible with httpbin. We probably need to add a new test server for such tests.

plugins/modules/get_certificate.py
```diff
@@ -18,7 +18,7 @@
       library. By default, it tries to detect which one is available. This can be
       overridden with the I(select_crypto_backend) option. Please note that the PyOpenSSL
       backend was deprecated in Ansible 2.9 and will be removed in community.crypto 2.0.0."
-    - Support SNI only with python >= 2.7
+    - Support SNI only with python >= 2.7 (L(Server Name Indication,https://en.wikipedia.org/wiki/Server_Name_Indication))
```
felixfontein, line 21:
```suggestion
    - Support SNI (L(Server Name Indication,https://en.wikipedia.org/wiki/Server_Name_Indication)) only with python >= 2.7.
```

plugins/modules/get_certificate.py
```diff
@@ -297,7 +305,10 @@ def main():
                 ctx.check_hostname = False
                 ctx.verify_mode = CERT_NONE
 
-            cert = ctx.wrap_socket(sock, server_hostname=host).getpeercert(True)
+            if not server_name:
+                server_name = host
+
```
felixfontein, line 310: With the change in the next line, these lines are no longer needed.
```suggestion
```

felixfontein (approved):

NorthFuture:
> Please note the failing sanity tests:
> 
> ```
> plugins/modules/get_certificate.py:308:0: trailing-whitespace: Trailing whitespace
> ```
> 
> About testing: I'm not sure probably testing this is possible with httpbin. We probably need to add a new test server for such tests.

Fixed sanity tests. I'll take a look at the integration test. I'm not a regular ansible developer, nor python developer, so I'm a bit clumsy

felixfontein:
@NorthFuture thanks! I also don't mind merging this without tests (the code change is simple enough), but if you can improve the tests that would be even more awesome!

NorthFuture:
> @NorthFuture thanks! I also don't mind merging this without tests (the code change is simple enough), but if you can improve the tests that would be even more awesome!

Luckily, the existing test server already serves two different certificates based on server name  identification. Added two tests.

felixfontein (approved):

felixfontein:
@NorthFuture that's great to know, thanks!

felixfontein:
BTW, some of the commits have (potentially) strange 'co-authored-by' values. Please check and modify them if you want them gone. If you don't mind, tell me, then I'll merge (we always squash-merge, and these will end up in the commit message).
```
Co-authored-by: root <jenkinsops@previdom.previnet.it>
Co-authored-by: Prevedello Cristian <root@srvdockertest01.previdom.previnet.it>
```

NorthFuture:
> BTW, some of the commits have (potentially) strange 'co-authored-by' values. Please check and modify them if you want them gone. If you don't mind, tell me, then I'll merge as-is (we always squash-merge, and these will end up in the commit message).
> 

Doh, I jumped on different servers and containers to test, integrate and and  develop and I haven't realized git configurations were wrong. I'll remove them. Could you remove your message as well?

felixfontein:
I completely deleted it since I cannot delete the edit revisions.

felixfontein (approved):

felixfontein:
@NorthFuture thanks for implementing this!

PR #175, opened 2021-01-22
acme_certificate: Improve challenge docs

idawko:
##### SUMMARY

Some of the challenge docs are not fully functional because the playbook fails with:
```
The task includes an option with an undefined variable. The error was: 'list object' has no attribute 'value'
```

or

```
The task includes an option with an undefined variable. The error was: 'list object' has no attribute 'key'
```

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME
acme_certificate

##### ADDITIONAL INFORMATION

`dict2items` filter should be used when accessing `item.key` and `item.value` with `loop` keyword as documented [here](https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#with-dict) (option 1)


felixfontein (approved):
Good spot, thanks!

felixfontein:
@idawko thanks for improving the docs!

PR #180, opened 2021-01-29
openssl_csr: handle missing basic constraint

### LINKED ISSUE #179: openssl_csr: fails if existing CSR has no basic constraint but new one does
##### SUMMARY

If there is an existing CSR that does not include a basic constraint but the new one does openssl_csr tries to retrieve the critical flag resulting in 'NoneType' object has no attribute 'critical'. 

Unexpected corner case caused by mistyping basic_contraint;)

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

openssl_csr
##### ANSIBLE VERSION

```paste below

```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

schallee:
##### SUMMARY

Fixes #179 by verifying the basic constraint exists before checking it is critical.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

openssl_csr
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/179-openssl-csr-basic-constraint.yml
+++ b/changelogs/fragments/179-openssl-csr-basic-constraint.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - "openssl_csr - fails when comparing CSR without basic constraint"
+  - "openssl_csr - no longer fails when comparing CSR without basic constraint when ``basic_constraints`` is specified (https://github.com/ansible-collections/community.crypto/issues/179, https://github.com/ansible-collections/community.crypto/pull/180)."
--- a/changelogs/fragments/179-openssl-csr-basic-constraint.yml
+++ b/changelogs/fragments/179-openssl-csr-basic-constraint.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - "openssl_csr - fails when comparing CSR without basic constraint"
--- a/plugins/module_utils/crypto/module_backends/csr.py
+++ b/plugins/module_utils/crypto/module_backends/csr.py
@@ -626,11 +626,9 @@ class CertificateSigningRequestCryptographyBackend(CertificateSigningRequestBack
                 return False
             # Check criticality
             if self.basicConstraints:
-                if not bc_ext or bc_ext.critical != self.basicConstraints_critical:
-                    return False
+                return bc_ext is not None and bc_ext.critical == self.basicConstraints_critical
             else:
                 return bc_ext is None
-            return True
 
         def _check_ocspMustStaple(extensions):
             try:
--- a/plugins/module_utils/crypto/module_backends/csr.py
+++ b/plugins/module_utils/crypto/module_backends/csr.py
@@ -626,8 +626,10 @@ class CertificateSigningRequestCryptographyBackend(CertificateSigningRequestBack
                 return False
             # Check criticality
             if self.basicConstraints:
-                if bc_ext.critical != self.basicConstraints_critical:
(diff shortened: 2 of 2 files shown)


felixfontein:
Thanks for the PR! I'll take a look tomorrow.

plugins/module_utils/crypto/module_backends/csr.py
```diff
@@ -626,8 +626,10 @@ def _check_basicConstraints(extensions):
                 return False
             # Check criticality
             if self.basicConstraints:
-                if bc_ext.critical != self.basicConstraints_critical:
+                if not bc_ext or bc_ext.critical != self.basicConstraints_critical:
```
felixfontein, line 629: How about inverting the logic and doing:
```suggestion
                return bc_ext is not None and bc_ext.critical == self.basicConstraints_critical
```
and removing the `return False` / `return True` below? I think that makes the code easier to read and the conditions easier to understand.

felixfontein (commented):
Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

schallee:
I have made the suggest modifications.

I frankly find the logic in this method confusing. Initially I rewrote the whole method but decided on a more minimal change. The logic seems to be get the fields from both objects, giving them non-matching defaults if missing and then compare the fields one at a time. I would have tested that both objects existed before doing so which would caused this bug to never happen. I also tend to check each field for equality and then return true after all checks letting the optimizer convert the return true to the result of the last comparison because it makes it far easier to modify in the future. This is mostly taste and style which is why I went with the minimal change.

Hopefully the change log fragment is sufficient. There is only one other in this repo for an example;)

changelogs/fragments/179-openssl-csr-basic-constraint.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - "openssl_csr - fails when comparing CSR without basic constraint"
```
felixfontein, line 3:
```suggestion
  - "openssl_csr - no longer fails when comparing CSR without basic constraint when ``basic_constraints`` is specified (https://github.com/ansible-collections/community.crypto/issues/179, https://github.com/ansible-collections/community.crypto/pull/180)."
```

felixfontein:
> I have made the suggest modifications.

Thanks!

> I frankly find the logic in this method confusing. Initially I rewrote the whole method but decided on a more minimal change.

I agree, I cannot remember why the function is as it is now, and I don't think there's a reason.

> The logic seems to be get the fields from both objects, giving them non-matching defaults if missing

What do you mean by "non-matching defaults"? `False` for CA and `None` for path length are consistent with `cryptography_get_basic_constraints()`, and are the values that make most sense for a default (not a CA, and there is no restriction on path length).

> Hopefully the change log fragment is sufficient. There is only one other in this repo for an example;)

For every release, all fragments are incorporated in changelogs/changelog.yaml, so you can find more examples there: https://github.com/ansible-collections/community.crypto/blob/main/changelogs/changelog.yaml#L289

schallee:
Oops, I meant in the general case not the specific case. Eg: class has fields field1 and field2, if object_a.field1 != object_b.field1 return false. If object_a.field2 != object.field2 return false. Finally, return true;)

felixfontein (approved):

felixfontein:
@schallee thanks for fixing this!

PR #185, opened 2021-02-05
exclude test folder from release

schurzi:
##### SUMMARY
Don't include tests and pipeline definition in galaxy release, since these only increase package size and do not bring any benefit.

##### ISSUE TYPE
- Bugfix Pull Request ?
- Feature Pull Request ?

##### ADDITIONAL INFORMATION
currently following files are included in the release:
```
$ ls -la
total 196
drwxr-xr-x 7 root root   4096 Feb  5 18:50 .
drwxr-xr-x 3 root root   4096 Feb  5 17:00 ..
drwxr-xr-x 4 root root   4096 Feb  5 17:00 .azure-pipelines
-rw-r--r-- 1 root root  15691 Feb  5 17:00 CHANGELOG.rst
-rw-r--r-- 1 root root  35149 Feb  5 17:00 COPYING
-rw-r--r-- 1 root root 102410 Feb  5 17:00 FILES.json
-rw-r--r-- 1 root root    997 Feb  5 17:00 MANIFEST.json
-rw-r--r-- 1 root root   5459 Feb  5 17:00 README.md
drwxr-xr-x 3 root root   4096 Feb  5 18:51 changelogs
drwxr-xr-x 2 root root   4096 Feb  5 17:00 meta
drwxr-xr-x 7 root root   4096 Feb  5 17:00 plugins
drwxr-xr-x 6 root root   4096 Feb  5 17:00 tests
```

Should we also exclude the `changelogs` folder?

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -25,6 +25,8 @@ documentation: https://docs.ansible.com/ansible/latest/collections/community/cry
 homepage: https://github.com/ansible-collections/community.crypto
 issues: https://github.com/ansible-collections/community.crypto/issues
 build_ignore:
+  - .azure-pipelines
   - 'community-crypto-*.tar.gz'
   - .gitignore
   - changelogs/.plugin-cache.yaml
+  - tests


felixfontein (changes requested):
So far the policy for community collections is to contain all material used to build them, and that includes test and CI config. (There are potential legal problems when we don't do that.)

Also, simply excluding all of `tests/` is a Very Bad Idea, since this also excludes `tests/sanity/`, which makes it impossible to run sanity tests on the installed collection without getting new errors.

schurzi:
thank you for clarifying that. I will close this now, as there is nothing left to be done.

PR #193, opened 2021-02-28
luks_device: add sector_size option

### LINKED ISSUE #192: luks_device: Support sector size for luksFormat
##### SUMMARY
LUKS2 allows for using sector sizes larger than 512 bytes. This has performance advantages. 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
luks_device

##### ADDITIONAL INFORMATION
Add an option, sector_size to luks_device. On formatting, this would use the --sector-size parameter to use a different sector size or fail when luks2 isn't being used. An additional challenge may be that luksFormat will only format a device if the device size dividable by the sector size, users will have to bear this in mind when dealing with underlying devices. 

```yaml
- name: Create LUKS container with a passphrase and 4096 bytes sector size
  community.crypto.luks_device:
    device: "/dev/loop0"
    state: "present"
    sector_size: 4096
    type: luks2
    passphrase: "foo"

```

felixfontein:
##### SUMMARY
Fixes #192.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
luks_device

### DIFF
--- a/changelogs/fragments/193-luks_device-sector_size.yml
+++ b/changelogs/fragments/193-luks_device-sector_size.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- "luks_device - allow to specify sector size for LUKS2 containers with new ``sector_size`` parameter (https://github.com/ansible-collections/community.crypto/pull/193)."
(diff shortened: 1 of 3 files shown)


briantist (approved):
I'm not too familiar with this module but this change looks straightforward to me.

felixfontein:
@briantist thanks for reviewing this! :)

Issue #107, opened 2020-09-07
 openssh_keypair error if key already exists


devnewton:
##### SUMMARY

openssh_keypair throws an error if ssh key pair already exists.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
openssh_keypair module

##### ANSIBLE VERSION

```paste below
ansible 2.9.13
  configured module search path = ['/home/devnewton/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/devnewton/.local/lib/python3.8/site-packages/ansible
  executable location = /home/devnewton/.local/bin/ansible
  python version = 3.8.3 (default, May 23 2020, 15:50:53) [GCC 9.3.0]
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT
Windows 10 + cygwin

##### STEPS TO REPRODUCE
Run a playbook with openssh_keypair twice.

```yaml
- name: test ssh key gen
  hosts: localhost
  gather_facts: false
  tasks:
    - name: create target directory
      file:
        path: "target"
        state: directory
    - name: create ssh key
      openssh_keypair:
        path: "target/ssh_key" 
```

##### EXPECTED RESULTS
No error expected.

##### ACTUAL RESULTS
Playbook fails on python exception.

```paste below
An exception occurred during task execution. To see the full traceback, use -vvv. The error was:   File "/tmp/ansible_openssh_keypair_payload_rzdh605t/ansible_openssh_keypair_payload.zip/ansible/modules/crypto/openssh_keypair.py", line 292, in _pubkey_valid
fatal: [localhost]: FAILED! => {
    "changed": false
}

MSG:

'bool' object is not subscriptable
```

felixfontein:
I cannot reproduce this with your playbook, neither for stable-2.9 nor devel branch of Ansible. But then, I neither have Windows 10 nor cygwin running, so it could well be related to that.

srgg:
I have the same issue:

```
ansible --version                                                            2 ↵  2915  17:31:21
ansible 2.9.13
  config file = /Users/user/work/../ansible.cfg
  configured module search path = ['/Users/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
  executable location = /usr/local/bin/ansible
[...]

PR #196, opened 2021-03-14
Adding support for passphrase with openssh_keypair

Ajpantuso:
##### SUMMARY
Allowing users to generate a keypair with a supplied passphrase. (And consequently manage private keys with passphrases)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/openssh_keypair.py

##### ADDITIONAL INFORMATION
Most of the work for this was already in the module with the updates being:
- New parameter for passphrase
- Updates to any invocations of ssh-keygen to use either the -N or -P flags when passphrase is required.
- An additional helper function to determine when the user has supplied a passphrase if it was previously null (Providing a passphrase when the key is unprotected does not produce an error so this was necessary)
- Use of the new helper function to maintain expected behavior when using various idempotency options.
- Updates to documentation for new parameter however the regenerate parameter description already defined the behavior for how the module handles idempotency with passphrases. (Even though it was not implemented)

### DIFF
--- a/plugins/modules/openssh_keypair.py
+++ b/plugins/modules/openssh_keypair.py
@@ -185,7 +185,6 @@ class Keypair(object):
         self.fingerprint = {}
         self.public_key = {}
         self.regenerate = module.params['regenerate']
-
         if self.regenerate == 'always':
             self.force = True
 
--- a/plugins/modules/openssh_keypair.py
+++ b/plugins/modules/openssh_keypair.py
@@ -41,6 +41,14 @@ options:
         type: str
         default: rsa
         choices: ['rsa', 'dsa', 'rsa1', 'ecdsa', 'ed25519']
+    passphrase:
+        description:
+            - An optional passphrase to secure the newly generated private key.
+            - If a passphrase has already been applied to a private key, then this parameter must be
+              provided with the same passphrase to consider the private key valid.
+            - See the C(regenerate) parameter for further clarification on when passphrase changes
+              will result in a newly generated key-pair or task failure.
+        type: str
     force:
         description:
             - Should the key be regenerated even if it already exists
@@ -177,8 +185,14 @@ class Keypair(object):
         self.fingerprint = {}
         self.public_key = {}
         self.regenerate = module.params['regenerate']
+
         if self.regenerate == 'always':
             self.force = True
+
+        if module.params['passphrase']:
+            self.passphrase = module.params['passphrase']
+        else:
+            self.passphrase = ''
 
         if self.type in ('rsa', 'rsa1'):
             self.size = 4096 if self.size is None else self.size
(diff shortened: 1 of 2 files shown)


plugins/modules/openssh_keypair.py
```diff
@@ -207,7 +220,7 @@ def generate(self, module):
             args = [
                 module.get_bin_path('ssh-keygen', True),
                 '-q',
-                '-N', '',
+                '-N', self.passphrase,
```
felixfontein, line 223: This won't be merged as-is. Passprases must **never** be passed as command line parameter, they must always be passed via stdin.

(Also, you might be interested in #107, in particular: https://github.com/ansible-collections/community.crypto/issues/107#issuecomment-761848183 Also there's https://github.com/ansible/ansible/pull/46436#discussion_r222757521 Just so you know what kind of can of worms you are about to open ;-) )

Ajpantuso: Lol, whoops.
Definitely don't write your secrets to the proc table. (Should've caught that)
Yeah, this module should probably be left alone unless someone wants to take inspiration from [builtin.user](https://github.com/ansible/ansible/blob/9ec4e08534209a2076023489c4a3f50f464d3b7a/lib/ansible/modules/user.py#L1148)
An additional module with a limited feature set not dependent on ssh-keygen could also be done using something like [this](https://pycryptodome.readthedocs.io/en/latest/src/public_key/rsa.html#Crypto.PublicKey.RSA.RsaKey.exportKey)
Thanks for the context and hopefully this will clarify the issue for anyone who is seeking this feature.

Ajpantuso:
This is a non-trivial feature and would require significantly more effort than what was initially scoped.
See review and reply for more details

plugins/modules/openssh_keypair.py
felixfontein: Yes, avoiding ssh-keygen would be the best solution... (also for other reasons, like the issues in #107)

An alternative would be to use cryptography; it can load private and public keys:
* https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization.html#openssh-public-key
* https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization.html#openssh-private-key

It can also serialize some things so that OpenSSH should understand them:
* https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization.html#cryptography.hazmat.primitives.serialization.PublicFormat.OpenSSH
* https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization.html#cryptography.hazmat.primitives.serialization.PrivateFormat.OpenSSH
* https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization.html#cryptography.hazmat.primitives.serialization.Encoding.OpenSSH

I've never used that parts of cryptography, so no idea how well it works, but in general cryptography is pretty good (also many modules in this collection use/support it).

Ajpantuso: Yep, I was looking at cryptography as well.
Ultimately a partial or full implementation of the ssh-keygen feature set in python should probably only come from a security engineer to avoid the rabbit hole of security concerns. 
Also seems a little broader in scope than this project.

PR #198, opened 2021-03-15
Improve openssl_privatekey docs: mention default mode in more places

felixfontein:
##### SUMMARY
I originally wanted to change the `default` of `mode` to `0600`, but that's more complicated the way `add_file_common_args=True` works (the argspec entries are only added in `AnsibleModule` itself, if they haven't been present, so I would have had to replace the complete `mode` entry - which I think is more dangerous than just changing its `default` value). So I've added one more remark about the default mode being `0600` to the main description (currently it's only mentioned in `path`'s description).

CC @bmillemathias 

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
openssl_privatekey

### DIFF
--- a/plugins/modules/openssl_privatekey.py
+++ b/plugins/modules/openssl_privatekey.py
@@ -14,6 +14,7 @@ module: openssl_privatekey
 short_description: Generate OpenSSL private keys
 description:
     - This module allows one to (re)generate OpenSSL private keys.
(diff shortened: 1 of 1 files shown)


bmillemathias (commented):
in the absence of possibility of having the text in `mode` this is good to have it here.

felixfontein:
@bmillemathias thanks for reviewing this!

PR #213, opened 2021-03-28
New module_utils openssh

Ajpantuso:
##### SUMMARY
Utilities for working with OpenSSH key pairs abstracted from `cryptography`.
The intention is to offer these utilities as an alternative to invoking ssh-keygen.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/module_utils/crypto/cryptography_openssh.py
tests/unit/plugins/module_utils/crypto/test_cryptography_openssh.py

##### ADDITIONAL INFORMATION
This exposes two classes, Asymmetric_Keypair for encapsulating asymmetric keypairs supported by `cryptography` and OpenSSH_Keypiar for the OpenSSH encoded/formatted representation of a keypair, mimicking some of the features offered by ssh-keygen.

Attributes of both classes for the most part are exposed as read-only properties to reinforce that keys are immutable.
However comments applied to OpenSSH formatted public keys can be updated.

Since ed25519 is only available from `cryptography` >=2.6 the utilities will only successfully import if that condition is met.
A soft requirement is `cryptography` >=3.0 for OpenSSH formatted private keys, but ssh-keygen is not consistent about formatting private keys this way depending on version either.

Text encoding is set to UTF-8 globally in the module and could be changed to ASCII if neccessary.

SHA256 is used as the standard hash algorithm to align with current ssh-keygen versions.

I removed a couple of unit tests for directly comparing ssh-keygen to `cryptography` generated keys and vise versa because I wasn't sure how best to implement command invocation outside of an Ansible Module. (Direct comparisons are positive in my local test environment for whatever that is worth)

Definitely looking for input on this one as this is for the most part just a straight encapsulation of the `cryptography` library with most of the effort spent on how to present it as a utility in this collection.


plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+__metaclass__ = type
+
+from base64 import b64encode, b64decode
+from getpass import getuser
+from socket import gethostname
+
+try:
+    from .basic import (
+        HAS_CRYPTOGRAPHY,
+        CRYPTOGRAPHY_HAS_ED25519,
```
felixfontein, line 28: These symbols are always defined, there's no need to catch an `ImportError` here.

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+
+if HAS_CRYPTOGRAPHY and CRYPTOGRAPHY_HAS_ED25519:
+
+    from cryptography.exceptions import UnsupportedAlgorithm
+    from cryptography.hazmat.backends.openssl import backend
+    from cryptography.hazmat.primitives import hashes, serialization
+    from cryptography.hazmat.primitives.asymmetric import dsa, ec, rsa, padding
+    from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
+
+    _TEXT_ENCODING = 'UTF-8'
```
felixfontein, line 43: I think this and the following should not be in the `if`. You should export a combined symbol `HAS_OPENSSH_SUPPORT` though which is `HAS_CRYPTOGRAPHY and CRYPTOGRAPHY_HAS_ED25519`, so that users can check that one (instead of having to know which exact flags from `crypto.basic` this module utils uses).

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+
+    class InvalidAlgorithmError(Exception):
+        pass
+
+    class InvalidCommentError(Exception):
```
felixfontein, line 48: How about having a base class for all these exceptions, like `OpenSSHError`?

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+
+                if size:
+                    self.__size = size
+                else:
+                    self.__size = self.__algorithm_parameters[self.__keytype]['default_size']
+
+                self.__generate_keypair()
+
+        @property
+        def PrivateKey(self):
```
felixfontein, line 131: Such symbols are reserved for type names in Python. The correct convention for properties is:
```suggestion
        def private_key(self):
```

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+            """Returns the openssh formatted private key of this key pair
```
felixfontein, line 132:
```suggestion
            """Returns the OpenSSH formatted private key of this key pair
```
(same in other comments)

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+                    pubkey_text = content.decode(encoding=_TEXT_ENCODING)
+
+            except ValueError as e:
+                raise InvalidKeyFileError(e)
+            except UnsupportedAlgorithm as e:
+                raise InvalidAlgorithmError(e)
+
+            return publickey
+
+        def __validate_passphrase(self, passphrase):
```
felixfontein, line 324: Why not make this function raise `InvalidPassphraseError` directly? That would simplify its usage.

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+    class InvalidKeyTypeError(Exception):
+        pass
+
+    class InvalidPassphraseError(Exception):
+        pass
+
+    class Asymmetric_Keypair(object):
+        """Container for newly generated asymmetric keypairs or those loaded from existing files
+
+           :path: A path to an existing private key to be loaded
```
felixfontein, line 69: These are not properties of the class.

If you want to document the constructor arguments, this must be in the docstring for the constructor.

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+                    raise InvalidKeyTypeError("%s is not a valid keytype. Valid keytypes are %s" % (keytype, ", ".join(self.__algorithm_parameters.keys())))
+
+                self.__keytype = keytype
+
+                if size:
+                    self.__size = size
+                else:
+                    self.__size = self.__algorithm_parameters[self.__keytype]['default_size']
+
+                self.__generate_keypair()
```
felixfontein, line 128: I think it would be better to have class methods with speaking names to do these tasks, i.e. `Asymmetric_Keypair.load()` to load a keypair and `Asymmetric_Keypair.generate()` to generate one.

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+                signed_data = self.__privatekey.sign(
+                    data,
+                    **signer_params[self.__keytype]
+                )
+            except TypeError as e:
+                raise InvalidDataError(e)
+
+            return signed_data
+
+        def Update_Passphrase(self, passphrase):
```
felixfontein, line 192: Method names should follow the same convention as property names:
```suggestion
        def update_passphrase(self, passphrase):
```

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
...
+        def __validate_passphrase(self, passphrase):
+            err = ""
+
+            if not hasattr(passphrase, 'encode'):
+                err = "%s cannot be encoded to text" % passphrase
+
+            return err
+
+    class OpenSSH_Keypair(object):
+        """Container for newly generated asymmetric keypairs or those loaded from existing files
```
felixfontein, line 333: What's the difference between `Asymmetric_Keypair` and this class? Both descriptions are identical.

plugins/module_utils/crypto/cryptography_openssh.py
```diff
@@ -0,0 +1,484 @@
+# -*- coding: utf-8 -*-
```
felixfontein, line 1: About the placement of this file: I think it would be better to have a separate directory, `plugins/module_utils/openssh/`, instead of cramming more content into `plugins/module_utils/crypto/`.

Ajpantuso: Moved the file to plugins/module_utils/crypto and incorporated the remaining suggestions from your review.
Thank you.

plugins/module_utils/openssh/cryptography_openssh.py
```diff
@@ -0,0 +1,490 @@
...
+    pass
+
+
+class InvalidPassphraseError(OpenSSHError):
+    pass
+
+
+class Asymmetric_Keypair(object):
+    def __init__(self, path=None, keytype='rsa', size=None, passphrase=None):
+        """Container for newly generated asymmetric keypairs or those loaded from existing files
```
felixfontein, line 85: This line should be part of the class's docstring, not of the `__init__` docstring. It describes what the class does, not what `__init__` does. (Same for the other class.)

plugins/module_utils/crypto/cryptography_openssh.py
felixfontein: Have you thought about this one?

Ajpantuso: I did remove _keypair from these method names, but I may be missing the meaning of your suggestion.
Was the intention that these be readable as _actions_ solely in the context of initialization or that the methods also be exposed for end users? I intentionally wanted to present asymmetric keypairs as immutable single-use objects hence why the loading/generating mechanisms are obscured.

felixfontein: I would offer them as class methods, i.e. you can do `AsymmetricKeypair.load(filename)` to load a keypair, which will return an `AsymmetricKeypair` instance. This is more expressive then using `AsymmetricKeypair(path=filename)` IMO.

felixfontein: (See for example https://code-maven.com/slides/python/class-methods-alternative-constructor for this kind of pattern.)

Ajpantuso: Ok, that does open the possibility of an _empty_ object and makes the faux overloaded constructor less confusing. I will spend some time on that. Thanks

felixfontein: You could declare the constructor private (by docstring, no idea whether Python allows to do that explicitly without using ugly hacks) to avoid having an empty keypair.

Ajpantuso: I am considering the possibility of an empty object a benefit.
That would allow for the truthiness of the class to be evaluated in a more obvious (pythonic) way.

felixfontein: Hmm, I think it's a lot easier to let only `None` represent an empty key, instead of having an instance of the class which can represent an empty key. While such an empty value could simplify some cases, it in general makes code more complicated, since it is harder to ensure that a key is actually specified. (For example, it is basically impossible to specify this requirement with the type system, which is IMO a huge disadvantage.)

plugins/module_utils/openssh/cryptography_openssh.py
Ajpantuso: Ok, generate/load are now exposed as class methods and docstrings are cleaned up.
Updated unit tests to use generate/load as well just for illustration.

felixfontein:
The change c22773321cc28cade2cce80af6d80bae83a847b5 is not what I meant. It basically reverts the PR to what it was at the beginning (with the class methods thin wrappers around `__init__`).

The `__init__` function should only look like this:
```
    def __init__(self, keytype, privatekey, publickey, size):
        """
           :keytype: One of rsa, dsa, ecdsa, ed25519
           :privatekey: The cryptography pivate key class
           :publickey: The cryptography public key class
           :size: The key size
        """
        self.__encryption_algorithm = serialization.NoEncryption()
        self.__keytype = keytype
        self.__privatekey = privatekey
        self.__publickey = publickey
        self.__size = size
```

The 'hard work' should be done by `Asymmetric_Keypair.generate` and `Asymmetric_Keypair.load`. They do the work, and once done call `__init__` to hand over the generated / loaded data.

Ajpantuso:
> The change [c227733](https://github.com/ansible-collections/community.crypto/commit/c22773321cc28cade2cce80af6d80bae83a847b5) is not what I meant. It basically reverts the PR to what it was at the beginning (with the class methods thin wrappers around `__init__`).
> 
> The `__init__` function should only look like this:
> 
> ```
>     def __init__(self, keytype, privatekey, publickey, size):
>         """
>            :keytype: One of rsa, dsa, ecdsa, ed25519
>            :privatekey: The cryptography pivate key class
>            :publickey: The cryptography public key class
>            :size: The key size
>         """
>         self.__encryption_algorithm = serialization.NoEncryption()
>         self.__keytype = keytype
>         self.__privatekey = privatekey
>         self.__publickey = publickey
>         self.__size = size
> ```
> 
> The 'hard work' should be done by `Asymmetric_Keypair.generate` and `Asymmetric_Keypair.load`. They do the work, and once done call `__init__` to hand over the generated / loaded data.

I was avoiding that because it required re-scoping the helper/utility functions, but the changes are incorporated as of the latest commit.

As a side note, how do I clear this sanity check?
`ERROR: plugins/module_utils/openssh/cryptography_openssh.py:87:0: traceback: NameError: name 'padding' is not defined`

I didn't change how `padding` is imported, but previously it was referred to in an instance variable and now it's being referred to in a class variable.

Also the affected sign method which requires `padding` to be defined works as expected.

felixfontein:
`/rebuild_failed` doesn't work here (since we don't have a bot here), but I've restarted them manually :)

plugins/module_utils/openssh/cryptography_openssh.py
```diff
@@ -0,0 +1,604 @@
...
+        size = privatekey.key_size
+
+        if isinstance(privatekey, rsa.RSAPrivateKey):
+            keytype = 'rsa'
+        elif isinstance(privatekey, dsa.DSAPrivateKey):
+            keytype = 'dsa'
+        elif isinstance(privatekey, ec.EllipticCurvePrivateKey):
+            keytype = 'ecdsa'
+        elif isinstance(privatekey, Ed25519PrivateKey):
+            keytype = 'ed25519'
```
felixfontein, line 221: This needs a final `else` block. It should either raise an exception, or set `keytype` to a concrete value.

plugins/module_utils/openssh/cryptography_openssh.py
```diff
@@ -0,0 +1,604 @@
...
+
+        return signature
+
+    def verify(self, signature, data):
+        """Verifies that the signature associated with the provided data was signed
+           by the private key of this key pair.
+
+           :signature: signature to verify
+           :data: byteslike data signed by the provided signature
+        """
```
felixfontein, line 303: I think it is a better API to let `verify()` raise an exception in case verification fails, and return nothing in case validation succeeds, i.e. do the same thing as cryptography also does.

plugins/module_utils/openssh/cryptography_openssh.py
```diff
@@ -0,0 +1,604 @@
...
+        return cls(
+            asym_keypair=asym_keypair,
+            openssh_privatekey=openssh_privatekey,
+            openssh_publickey=openssh_publickey,
+            fingerprint=fingerprint,
+            comment=comment
+        )
+
+    @classmethod
+    def encode_openssh_privatekey(cls, asym_keypair):
```
felixfontein, line 381:
```suggestion
    @staticmethod
    def encode_openssh_privatekey(asym_keypair):
```

plugins/module_utils/openssh/cryptography_openssh.py
```diff
@@ -0,0 +1,604 @@
...
+        encoded_privatekey = asym_keypair.private_key.private_bytes(
+            encoding=serialization.Encoding.PEM,
+            format=privatekey_format,
+            encryption_algorithm=asym_keypair.encryption_algorithm
+        )
+
+        return encoded_privatekey
+
+    @classmethod
+    def encode_openssh_publickey(cls, asym_keypair, comment):
```
felixfontein, line 402:
```suggestion
    @staticmethod
    def encode_openssh_publickey(asym_keypair, comment):
```

plugins/module_utils/openssh/cryptography_openssh.py
```diff
@@ -0,0 +1,604 @@
...
+        """
+
+        self.__asym_keypair = asym_keypair
+        self.__openssh_privatekey = openssh_privatekey
+        self.__openssh_publickey = openssh_publickey
+        self.__fingerprint = fingerprint
+        self.__comment = comment
+
+    def __eq__(self, other):
+        return self.fingerprint == other.fingerprint and self.comment == other.comment
```
felixfontein, line 435: I would rather compare the keys themselves, instead of only comparing the fingerprint. Or are some of the key objects not comparable?

(Also this ignores the passphrase.)

plugins/module_utils/openssh/cryptography_openssh.py
```diff
@@ -0,0 +1,604 @@
...
+        return self.__openssh_publickey
+
+    def update_passphrase(self, passphrase):
+        """Updates the passphrase used to encrypt the private key of this keypair
+
+           :passphrase: Text secret used for encryption
+        """
+
+        self.__asym_keypair.update_passphrase(passphrase)
+        self.__openssh_privatekey = type(self).encode_openssh_privatekey(self.__asym_keypair)
```
felixfontein, line 494:
```suggestion
        self.__openssh_privatekey = OpenSSH_Keypair.encode_openssh_privatekey(self.__asym_keypair)
```

felixfontein (commented):
Sorry it took so long to review this!

Ajpantuso:
> Sorry it took so long to review this!

No worries, this is definitely not a high priority and the reviews are very helpful.

Everything suggested makes sense and I will incorporate the changes.

Regarding the `__eq__` method it was definitely the lazy way to do it and yes ed25519 keys  have no implementation for comparison as far as I can tell (See [here](https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/primitives/asymmetric/ed25519.py)), but for the other key types it's simple to compare on the constituent components of the keys. (See [here](https://github.com/pyca/cryptography/blob/04ac16bce0380c0103548d69a542f8bda2795695/src/cryptography/hazmat/primitives/asymmetric/rsa.py#L372-L384) and [here](https://github.com/pyca/cryptography/blob/04ac16bce0380c0103548d69a542f8bda2795695/src/cryptography/hazmat/primitives/asymmetric/rsa.py#L372-L384) for examples with RSA) 

I can investigate more on the special case of ed25519 and probably just ask over on the cryptography side if they have any suggestions.

Thanks.

felixfontein:
Ah, that reminds me that I already had to solve that problem once ;-) See `cryptography_compare_public_keys` in `plugins/module_utils/crypto/cryptography_support.py`. I think comparing public keys suffices as well, since different private keys should not result in the same public key.

Ajpantuso:
> Ah, that reminds me that I already had to solve that problem once ;-) See `cryptography_compare_public_keys` in `plugins/module_utils/crypto/cryptography_support.py`. I think comparing public keys suffices as well, since different private keys should not result in the same public key.

Thanks, I made use of your example.
Seems like CI couldn't pull 'coverage' again, but otherwise I think CI should clear.
Besides the suggested changes I also:
- Stopped encoding `passphrase` input to bytes to allow the existing validation on the `Cryptography` side to run naturally
- Fixed a bug for loading ed25519 keys caused by the private key class not having a `key_size` attribute
- Cleaned up some comments
- Added an explicit test for various comparison scenarios

felixfontein (approved):
Let's get this merged! :tada:

felixfontein:
@Ajpantuso thanks for contributing this!

Ajpantuso:
> Let's get this merged! :tada:

Thanks for your help!

PR #225, opened 2021-05-06
openssh_keypair: Adding passphrase parameter

Ajpantuso:
##### SUMMARY
Added `passphrase` paramter to `openssh_keypair` for encrypting/decrypting OpenSSH private keys.
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME
plugins/modules/openssh_keypair
plugins/module_utils/openssh/cryptography_openssh
##### ADDITIONAL INFORMATION 
`openssh_keypair`
- Uses `cryptography` instead of _ssh_keygen_ to avoid passing passphrases over command line
- Maintains consistent formatting with _ssh_keygen_ which changed in version 7.8
- Ensures idempotency checks behave as documented when using passphrase
- Return values are consistent with parsed _ssh_keygen_ output
- `passphrase` is used as a switch so if empty or not supplied behavior will be the same as before this feature (Some more lines are included in `try` blocks, but are really just variable assignments)

`cryptography_openssh`
- Exposes SSH private key formatting capability as a global
- Added option to load a lone private key and automatically generate the corresponding public key
- Added option to explicitly select private key format
- Added logic to reattempt loading a PEM formatted private key if the OpenSSH loader fails
- Made it possible to set null comments
- Added "SHA256:" prefix to fingerprint

Integration tests have been updated as well.

Unfortunately the following containers have system packages of `cryptography` less than what is required to replicate the installed OpenSSH versions and the local pip versions are not able to install an updated version.

For these containers currently only the test cases which do not use the `passphrase` option are executed.
- CentOS 6
- CentOS 8
- openSUSE 15 py3
- Ubuntu 18.04
- RHEL 8.3
- FreeBSD 12.1
- FreeBSD 12.2
- FreeBSD 13.0

### DIFF
--- a/plugins/modules/openssh_keypair.py
+++ b/plugins/modules/openssh_keypair.py
@@ -7,7 +7,6 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-
 DOCUMENTATION = '''
 ---
 module: openssh_keypair
@@ -60,8 +59,17 @@ options:
     passphrase:
         description:
             - Passphrase used to decrypt an existing private key or encrypt a newly generated private key.
-            - Passphrases are not supported for I(type=rsa1)
+            - Passphrases are not supported for I(type=rsa1).
         type: str
+        version_added: 1.7.0
+    private_key_format:
+        description:
+            - Used when a value for I(passphrase) is provided to select a format for the private key at the provided I(path).
+            - The only valid option currently is C(auto) which will match the key format of the installed OpenSSH version.
+            - For OpenSSH < 7.8 private keys will be in PKCS1 format except ed25519 keys which will be in OpenSSH format.
+            - For OpenSSH >= 7.8 all private key types will be in the OpenSSH format.
+        type: str
+        default: auto
         version_added: 1.7.0
     regenerate:
         description:
@@ -217,18 +225,36 @@ class Keypair(object):
             ssh = module.get_bin_path('ssh', True)
             proc = module.run_command([ssh, '-Vq'])
             ssh_version = parse_openssh_version(proc[2].strip())
-            if LooseVersion(ssh_version) >= LooseVersion("7.8") and not HAS_OPENSSH_PRIVATE_FORMAT:
-                module.fail_json(
-                    msg=missing_required_lib(
-                        'cryptography >= 3.0',
-                        reason="to load/dump private keys in the default OpenSSH format for OpenSSH >= 7.8"
+
+            if module.params['private_key_format'] == 'auto':
+                if LooseVersion(ssh_version) >= LooseVersion("7.8") and not HAS_OPENSSH_PRIVATE_FORMAT:
+                    module.fail_json(
+                        msg=missing_required_lib(
+                            'cryptography >= 3.0',
+                            reason="to load/dump private keys in the default OpenSSH format for OpenSSH >= 7.8"
+                        )
                     )
-                )
-            elif LooseVersion(ssh_version) < LooseVersion("7.8"):
-                # OpenSSH made SSH formatted private keys available in version 6.5, but still defaulted to PEM format
-                self.private_key_format = 'PEM'
+                elif LooseVersion(ssh_version) < LooseVersion("7.8"):
+                    if self.type == 'ed25519':
+                        if HAS_OPENSSH_PRIVATE_FORMAT:
+                            # ed25519 keys are always formatted in the OpenSSH format by ssh-keygen
+                            # since their introduction in OpenSSH 6.5
+                            self.private_key_format = 'SSH'
+                        else:
+                            module.fail_json(
+                                msg=missing_required_lib(
+                                    'cryptography >= 3.0',
+                                    reason="to load/dump ed25519 private keys in the default OpenSSH format"
+                                )
+                            )
+                    else:
+                        # OpenSSH made SSH formatted private keys available in version 6.5,
+                        # but still defaulted to PKCS1 format
+                        self.private_key_format = 'PKCS1'
+                else:
+                    self.private_key_format = 'SSH'
             else:
-                self.private_key_format = 'SSH'
+                module.fail_json(msg="The only valid option for 'private_key_format' is 'auto'")
 
             if self.type == 'rsa1':
                 module.fail_json(msg="Passphrases are not supported for RSA1 keys.")
@@ -555,7 +581,8 @@ def main():
                 default='partial_idempotence',
                 choices=['never', 'fail', 'partial_idempotence', 'full_idempotence', 'always']
             ),
-            passphrase=dict(type='str', no_log=True)
+            passphrase=dict(type='str', no_log=True),
+            private_key_format=dict(type='str', default='auto')
         ),
         supports_check_mode=True,
         add_file_common_args=True,
--- a/plugins/modules/openssh_keypair.py
+++ b/plugins/modules/openssh_keypair.py
@@ -62,7 +62,7 @@ options:
             - Passphrases are not supported for I(type=rsa1).
         type: str
         version_added: 1.7.0
-    private_key_format:
+    key_format:
         description:
             - Used when a value for I(passphrase) is provided to select a format for the private key at the provided I(path).
             - The only valid option currently is C(auto) which will match the key format of the installed OpenSSH version.
@@ -226,7 +226,7 @@ class Keypair(object):
             proc = module.run_command([ssh, '-Vq'])
             ssh_version = parse_openssh_version(proc[2].strip())
 
-            if module.params['private_key_format'] == 'auto':
+            if module.params['key_format'] == 'auto':
                 if LooseVersion(ssh_version) >= LooseVersion("7.8") and not HAS_OPENSSH_PRIVATE_FORMAT:
                     module.fail_json(
                         msg=missing_required_lib(
@@ -239,7 +239,7 @@ class Keypair(object):
                         if HAS_OPENSSH_PRIVATE_FORMAT:
                             # ed25519 keys are always formatted in the OpenSSH format by ssh-keygen
                             # since their introduction in OpenSSH 6.5
-                            self.private_key_format = 'SSH'
+                            self.key_format = 'SSH'
                         else:
                             module.fail_json(
                                 msg=missing_required_lib(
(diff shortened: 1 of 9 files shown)


plugins/modules/openssh_keypair.py
```diff
@@ -180,6 +204,39 @@ def __init__(self, module):
...
+                    msg=missing_required_lib(
+                        'cryptography >= 2.6',
+                        reason="to encrypt/decrypt private keys with passphrases"
+                    )
+                )
+
+            ssh = module.get_bin_path('ssh', True)
+            proc = module.run_command([ssh, '-Vq'])
+            ssh_version = parse_openssh_version(proc[2].strip())
+            if LooseVersion(ssh_version) >= LooseVersion("7.8") and not HAS_OPENSSH_PRIVATE_FORMAT:
```
felixfontein, line 220: How about adding another option for determining the key format? It should also have an `auto` value (which would be the default, and do what exactly you are doing right now) which determines the format from `ssh -Vq`.

In general you want (for example) to be able to force the old format even if you have a newer SSH installed locally, since you plan to copy the key to a remote machine which has an older version of SSH. In that case, being able to specify the format is pretty important.

tests/integration/targets/openssh_keypair/tests/validate.yml
```diff
@@ -138,3 +138,34 @@
   assert:
     that:
       - privatekey8_result_force is changed
+
+- name: Check that password protected key with passphrase was regenerated
+  assert:
+    that:
+      - privatekey9_modified_result is changed
+  when: cryptography_version.stdout is version('3.0', '>=')
```
felixfontein, line 146: I would probably put all these asserts with the same `when:` into `block:` so you don't have to repeat the `when:` everywhere.

tests/integration/targets/openssh_keypair/tasks/setup_cryptography.yml
```diff
@@ -0,0 +1,33 @@
...
+# and should not be used as examples of how to write Ansible roles #
+####################################################################
+
+- name: Attempt to install dependencies for OpenSSH > 7.8
+  block:
+    - name: Ensure cryptography >= 3.0 available
+      become: true
+      pip:
+        name: cryptography>=3.0
+        extra_args: "-c {{ remote_constraints }}"
```
felixfontein, line 13: Why try to install `cryptography` when you added `setup_openssl` (which does exactly that) to `meta/main.yml`?

felixfontein (commented):
Thanks for this contribution! I really like it :)

What do you think of (probably more for a follow-up PR ;) ) adding backends to openssh_keypair, similar to the `cryptography` vs `pyopenssl` backends for some of the `openssl_*` modules? Assuming that the code for the `cryptography` backend would only rely on `cryptography` and no longer would need `ssh-keygen`.

tests/integration/targets/openssh_keypair/tasks/setup_cryptography.yml
Ajpantuso: Uh, I squashed all the ugly attempts at having CI clear, but the `setup_openssl` dependency, with the exception of Mac OS X, is only installing the system packaged cryptography version which is generally < 3.0 and in some cases < 2.6. I really only wanted the cryptography < 3.0 conditionals to take effect if absolutely necessary. Even with this hack the `pip` version on the following containers will not install a sufficient cryptography version: (I can check the logs again to collect which containers this task is actually `changed` for)

- CentOS 6
- CentOS 8
- openSUSE 15 py3
- Ubuntu 18.04
- RHEL 8.3
- FreeBSD 12.1
- FreeBSD 12.2
- FreeBSD 13.0

plugins/modules/openssh_keypair.py
Ajpantuso: I can do this, but I assume you would also like me to implement for the no `passphrase` scenario via `-m {PEM|SSH}`?
Just confirming before I take a crack at it.

Ajpantuso:
> Thanks for this contribution! I really like it :)
> 
> What do you think of (probably more for a follow-up PR ;) ) adding backends to openssh_keypair, similar to the `cryptography` vs `pyopenssl` backends for some of the `openssl_*` modules? Assuming that the code for the `cryptography` backend would only rely on `cryptography` and no longer would need `ssh-keygen`.

Freeing this module from `ssh-keygen` should be possible given it's current feature set.
I've researched other libraries while doing this and `cryptography` stands alone in terms of the OpenSSH specific features which are separate from the core `OpenSSL` library. And even then there are still gaps for example no support for generating SSH certificates which is critical for the `openssh_cert` module. If we maintained custom implementations of the OpenSSH formats that would open things up considerably....but then we would be stuck maintaining custom implementations of the OpenSSH formats :(

plugins/modules/openssh_keypair.py
felixfontein: I think for now I'd implement it only when `passphrase` is selected, and make it error out if set to any other value than `auto`. Implementing it for when ssh-keygen is used can be done in a follow-up PR (if you want, or someone else can do it when they want, or just implement it for the `cryptography` backend).

tests/integration/targets/openssh_keypair/tasks/setup_cryptography.yml
felixfontein: It is intentional that the other tests run with the OS-supplied `cryptography`, otherwise we would not need all these platform tests and could stick to the `default` container :) In fact, trying to install `cryptography` with a newer version here makes this very problematic for CI of the whole collection, since other tests running later might suddenly use another `cryptography` version than the one they used to.

felixfontein:
> Freeing this module from `ssh-keygen` should be possible given it's current feature set.

Great!

> I've researched other libraries while doing this and `cryptography` stands alone in terms of the OpenSSH specific features which are separate from the core `OpenSSL` library. And even then there are still gaps for example no support for generating SSH certificates which is critical for the `openssh_cert` module.

Well, I'd already be happy if only the `openssh_keypair` module has a `cryptography` backend and `openssh_cert` has not. I think there are a lot more people interested in `openssh_keypair` than there are in `openssh_cert`.

(`openssl_pkcs12` is another module without `cryptography` backend BTW, since PKCS12 support in cryptography is very, very basic right now, and only available since 3.0 IIRC.)

> If we maintained custom implementations of the OpenSSH formats that would open things up considerably....but then we would be stuck maintaining custom implementations of the OpenSSH formats :(

We definitely do **not** want to maintain that :-)

plugins/modules/openssh_keypair.py
Ajpantuso: Ok, I think everything is addressed, but I updated the 'PEM' format to be 'PKCS1' which I don't think is technically accurate, but fits the OpenSSH terminology. Cryptography just uses the name 'TraditionalOpenSSL' for the serialization format which uses the appropriate format for RSA, DSA, and ECDSA. Apparently ED25519 private keys are always in OpenSSH format even for newer versions of OpenSSH which support PKCS8 formatted keys so I added logic to check for this.

plugins/modules/openssh_keypair.py
```diff
@@ -19,6 +18,8 @@
       or C(ecdsa) private keys."
 requirements:
     - "ssh-keygen"
+    - cryptography >= 2.6 (if using C(passphrase) and OpenSSH < 7.8 is installed)
+    - cryptography >= 3.0 (if using C(passphrase) and OpenSSH >= 7.8 is installed)
```
felixfontein, line 22:
```suggestion
    - cryptography >= 2.6 (if using I(passphrase) and OpenSSH < 7.8 is installed)
    - cryptography >= 3.0 (if using I(passphrase) and OpenSSH >= 7.8 is installed)
```

plugins/modules/openssh_keypair.py
```diff
@@ -180,6 +212,57 @@ def __init__(self, module):
...
+                module.fail_json(
+                    msg=missing_required_lib(
+                        'cryptography >= 2.6',
+                        reason="to encrypt/decrypt private keys with passphrases"
+                    )
+                )
+
+            ssh = module.get_bin_path('ssh', True)
+            proc = module.run_command([ssh, '-Vq'])
+            ssh_version = parse_openssh_version(proc[2].strip())
```
felixfontein, line 227: I would move these lines into the following `if` block.

plugins/modules/openssh_keypair.py
```diff
@@ -55,6 +56,21 @@
...
+        type: str
+        version_added: 1.7.0
+    private_key_format:
+        description:
+            - Used when a value for I(passphrase) is provided to select a format for the private key at the provided I(path).
+            - The only valid option currently is C(auto) which will match the key format of the installed OpenSSH version.
+            - For OpenSSH < 7.8 private keys will be in PKCS1 format except ed25519 keys which will be in OpenSSH format.
+            - For OpenSSH >= 7.8 all private key types will be in the OpenSSH format.
+        type: str
+        default: auto
```
felixfontein, line 72:
```suggestion
        default: auto
        choices:
          - auto
```

plugins/modules/openssh_keypair.py
```diff
@@ -180,6 +212,57 @@ def __init__(self, module):
...
+                                    'cryptography >= 3.0',
+                                    reason="to load/dump ed25519 private keys in the default OpenSSH format"
+                                )
+                            )
+                    else:
+                        # OpenSSH made SSH formatted private keys available in version 6.5,
+                        # but still defaulted to PKCS1 format
+                        self.private_key_format = 'PKCS1'
+                else:
+                    self.private_key_format = 'SSH'
```
felixfontein, line 255: Isn't this basically equivalent to:
```suggestion
                self.private_key_format = 'SSH'
                if LooseVersion(ssh_version) < LooseVersion("7.8") and self.type != 'ed25519':
                    # OpenSSH made SSH formatted private keys available in version 6.5,
                    # but still defaulted to PKCS1 format
                    self.private_key_format = 'PKCS1'

                if self.private_key_format == 'SSH' and not HAS_OPENSSH_PRIVATE_FORMAT:
                    module.fail_json(
                        msg=missing_required_lib(
                            'cryptography >= 3.0',
                            reason="to load/dump private keys in the default OpenSSH format for OpenSSH >= 7.8 or for ed25519 keys"
                        )
                    )
```
?

plugins/modules/openssh_keypair.py
```diff
@@ -180,6 +212,57 @@ def __init__(self, module):
...
+            else:
+                module.fail_json(msg="The only valid option for 'private_key_format' is 'auto'")
```
felixfontein, line 257: With the `choices:`, this is no longer necessary.
```suggestion
```

plugins/modules/openssh_keypair.py
```diff
@@ -438,6 +581,8 @@ def main():
                 default='partial_idempotence',
                 choices=['never', 'fail', 'partial_idempotence', 'full_idempotence', 'always']
             ),
+            passphrase=dict(type='str', no_log=True),
+            private_key_format=dict(type='str', default='auto', no_log=False)
```
felixfontein, line 585:
```suggestion
            private_key_format=dict(type='str', default='auto', no_log=False, choices=['auto'])
```

tests/integration/targets/openssh_keypair/tasks/setup_bcrypt.yml
```diff
@@ -0,0 +1,24 @@
+---
```
felixfontein, line 1: This should be at some point moved to tests/integration/targets/setup_bcrypt/ so it can be used by other integration tests as well. No need to do that in this PR though :)

Ajpantuso: Yes, sorry was just pushing through issues with CI so there's definitely cleanup work to be done.

felixfontein (approved):
Tell me whether you want to do more work on the PR right now, or in a follow-up PR.

Ajpantuso:
> Tell me whether you want to do more work on the PR right now, or in a follow-up PR.

I'm good to go with this PR for now.

felixfontein:
@Ajpantuso thanks for working on this! :)

PR #234, opened 2021-05-18
openssl_pkcs12: add cryptography backend

### LINKED ISSUE #21: openssl_pkcs12: add cryptography backend
##### SUMMARY
The `openssl_pkcs12` module should support a cryptography backend, so PyOpenSSL can be deprecated / removed eventually. Currently cryptography does not support PKCS12, though, so some preliminary work on adding PKCS12 support to cryptography is needed. Also, this means that deprecating the PyOpenSSL backend would need to be done much later than for other modules, since a new enough cryptography supporting the new features (which aren't there yet) will take some years to trickle through.

(Migration of ansible/ansible#59905)

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
openssl_pkcs12

felixfontein:
##### SUMMARY
This requires cryptography 3.0 and @Ajpantuso's [hack](https://github.com/pyca/cryptography/issues/5760#issuecomment-842687238), so we can't deprecate pyOpenSSL anytime soon.

Also this is at the moment pretty ugly, I'll have to do a lot more cleanup...

Finally, `iter_size` and `maciter_size` are not supported. Also, the cryptography docs say `PKCS12 encryption is not secure and should not be used as a security mechanism. Wrap a PKCS12 blob in a more secure envelope if you need to store or send it safely. Encryption is provided for compatibility reasons only.`.

Fixes #21.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
openssl_pkcs12

### DIFF
--- a/plugins/modules/openssl_pkcs12.py
+++ b/plugins/modules/openssl_pkcs12.py
@@ -245,13 +245,13 @@ else:
     pyopenssl_found = True
 
 
-def load_certificate_set(filename):
+def load_certificate_set(filename, backend):
     '''
     Load list of concatenated PEM files, and return a list of parsed certificates.
     '''
     with open(filename, 'rb') as f:
         data = f.read().decode('utf-8')
-    return [load_certificate(None, content=cert) for cert in split_pem_list(data)]
+    return [load_certificate(None, content=cert, backend=backend) for cert in split_pem_list(data)]
 
 
 class PkcsError(OpenSSLObjectError):
@@ -267,6 +267,7 @@ class Pkcs(OpenSSLObject):
             module.params['force'],
             module.check_mode
         )
+        self.backend = 'pyopenssl'
         self.action = module.params['action']
         self.other_certificates = module.params['other_certificates']
         self.other_certificates_parse_all = module.params['other_certificates_parse_all']
@@ -293,10 +294,10 @@ class Pkcs(OpenSSLObject):
                 filenames = list(self.other_certificates)
                 self.other_certificates = []
                 for other_cert_bundle in filenames:
-                    self.other_certificates.extend(load_certificate_set(other_cert_bundle))
+                    self.other_certificates.extend(load_certificate_set(other_cert_bundle, self.backend))
             else:
                 self.other_certificates = [
-                    load_certificate(other_cert) for other_cert in self.other_certificates
+                    load_certificate(other_cert, backend=self.backend) for other_cert in self.other_certificates
                 ]
 
     def generate_bytes(self, module):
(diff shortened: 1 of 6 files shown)


plugins/modules/openssl_pkcs12.py
```diff
@@ -438,17 +518,152 @@ def parse(self):
...
+                ]
+
+            friendly_name = None
+            if certificate:
+                # See https://github.com/pyca/cryptography/issues/5760#issuecomment-842687238
+                friendly_name = certificate._backend._ffi.string(
+                    certificate._backend._lib.X509_alias_get0(
+                        certificate._x509, certificate._backend._ffi.new("int *")
+                    )
+                )
```
Ajpantuso, line 607:
```suggestion
                maybe_name = certificate._backend._lib.X509_alias_get0(
                    certificate._x509, certificate._backend._ffi.new("int *")
                )
                if maybe_name != certificate._backend._ffi.NULL:
                     friendly_name = certificate._backend._ffi.string(maybe_name)
```
Since this is a C function we need a null check (or try block) here otherwise if alias is `NULL` (cannot be retrieved from the x509 struct) we get a runtime error like `RuntimeError: cannot use string() on <cdata 'unsigned char *' NULL>`.

felixfontein: I was planning to look into that, but ran out of time yesterday and really had to go to bed ;-) Thanks for the suggestion, I've committed it.

felixfontein:
ready_for_review

tests/integration/targets/openssl_pkcs12/tasks/main.yml
```diff
@@ -4,6 +4,81 @@
...
+
+  - name: Generate CSRs
+    openssl_csr:
+      path: '{{ output_dir }}/ansible{{ item }}.csr'
+      privatekey_path: '{{ output_dir }}/ansible_pkey{{ item }}.pem'
+      commonName: www{{ item }}.ansible.com
+    loop:
+      - 1
+      - 2
+      - 3
```
Ajpantuso, line 32:
```suggestion
    loop: "{{ range(1, 4) }}"
```

plugins/modules/openssl_pkcs12.py
```diff
@@ -235,45 +268,62 @@
...
-def load_certificate_set(filename):
+
+def load_certificate_set(filename, backend):
     '''
     Load list of concatenated PEM files, and return a list of parsed certificates.
     '''
     with open(filename, 'rb') as f:
         data = f.read().decode('utf-8')
-    return [load_certificate(None, content=cert) for cert in split_pem_list(data)]
+    return [load_certificate(None, content=cert.encode('utf-8'), backend=backend) for cert in split_pem_list(data)]
```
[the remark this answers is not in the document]
felixfontein: PyOpenSSL has some magic to automatically encode: https://github.com/pyca/pyopenssl/blob/main/src/OpenSSL/crypto.py#L2898

tests/integration/targets/openssl_pkcs12/tasks/main.yml
```diff
@@ -4,6 +4,81 @@
...
-  when: pyopenssl_version.stdout is version('17.1.0', '>=')
+- block:
+  - name: Generate private keys
+    openssl_privatekey:
+      path: '{{ output_dir }}/ansible_pkey{{ item }}.pem'
+      size: '{{ default_rsa_key_size_certifiates }}'
+    loop:
+      - 1
+      - 2
+      - 3
```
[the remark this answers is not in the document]
felixfontein: Implemented in 044294fd3fbed164a23e842fb2a160c8a1a6431e.

tests/integration/targets/openssl_pkcs12/tasks/main.yml
felixfontein: Implemented in 044294fd3fbed164a23e842fb2a160c8a1a6431e.

Ajpantuso (approved):
LGTM, with the exception of CI for 2.9 which I broke by not remembering `range()` must be typecasted to list for compatibility. `"{{ range(1, 4) | list }}"` should clear the failures.

felixfontein:

PR #235, opened 2021-05-19
openssl_publickey_info: remove unnecessary and undocumented return values

felixfontein:
##### SUMMARY
The code fails anyway when not being able to load and parse public keys.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
openssl_publickey_info


felixfontein:
CC @Ajpantuso

Ajpantuso (approved):
Makes sense since these are just artifacts from when public key info gathering was embedded in the private key info module. LGTM.

felixfontein:

PR #236, opened 2021-05-22
openssh_keypair - Adding backend option and refactoring backend code

Ajpantuso:
##### SUMMARY
Adding `backend` option to `openssh_keypair` so that users can select between the OpenSSH binary and `cryptography`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/openssh_keypair

##### ADDITIONAL INFORMATION
- Backend code is modeled after other modules in this collection with two backends available to users.
- Some minor changes were made to the naming conventions in the openssh cryptography module_utils and a utility method also used by  `openssh_cert` was brought under the openssh directory to unify the library structure.
- Unit tests were updated to the new names in the module_utils and integration tests were enhanced to 1. move setup_bcrypt to it's own folder and 2. iterate through backends if the installed libraries of the test system are sufficient
- No intentional changes were made to the module's functionality (i.e. no bugfixes), but significant changes were made to the procedural structure, the control flow, and any unnecessary/unused variables/parameters
- Some cleanup to rationalize `try` blocks and use targeted exceptions was performed while also keeping calls to `fail_json` within the backend and having a single call to `exit_json` in the module `main` method
- Otherwise names were updated for improved readability for example `self.fingerprint` always contains the SHA256 fingerprint rather than sometimes the fingerprint and sometimes the full output of `ssh-keygen -lf FILE`


plugins/module_utils/openssh/backends/keypair_backend.py
```diff
@@ -0,0 +1,478 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright: (c) 2021, Andrew Pantuso (@ajpantuso) <ajpantuso@gmail.com>
```
felixfontein, line 3: Two quick comments on this file:
1. The reason other modules have most of their code in module_utils is that the code is used by multiple modules. Do you plan do add a new module that also uses this code? (If not, you this refactoring is also ok, just not really needed :) )
2. Since you probably took some code from the main module, you should include it's [copyright notice](https://github.com/ansible-collections/community.crypto/blob/83f01ab2fb88b959e125a3cc91b5aa15da3242f7/plugins/modules/openssh_keypair.py#L4) here as well.

plugins/modules/openssh_keypair.py
```diff
@@ -60,11 +60,12 @@
         description:
             - Passphrase used to decrypt an existing private key or encrypt a newly generated private key.
             - Passphrases are not supported for I(type=rsa1).
+            - Can only be used when C(backend=cryptography) or C(backend=auto) and the OpenSSH binary is not installed
```
felixfontein, line 63:
```suggestion
            - Can only be used when I(backend=cryptography), or when I(backend=auto) and a required C(cryptography) version is installed. 
```

felixfontein (commented):
Thanks for this PR! Some first quick comments:

plugins/module_utils/openssh/backends/keypair_backend.py
Ajpantuso: 1. No not in the immediate future (though I was considering some filters or lookups for OpenSSH public keys). I had originally planned on just passing the `backend` to the main class and toggling internally, but this module was clunky enough already. So if I wanted it to be more readable/maintainable I figured following a consistent pattern from the collection was probably a good start even if not necessary.
2. Just added; thanks for catching that.

plugins/modules/openssh_keypair.py
Ajpantuso: No matter how many times I check I still miss these, thanks. Will add a changelog fragment and open for review after I take a final look at this. And once this is ready for review I'll take some time to look at #150.

plugins/module_utils/openssh/utils.py
```diff
@@ -18,7 +18,6 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-
 import re
```
felixfontein, line 21: You need to keep an empty shell of this file in the old location which imports `parse_openssh_version` from the new version to avoid breaking backwards compatibility (module_utils is also part of the public API of the collection).

Ajpantuso: Added the original back with just the import in the latest commit.

tests/integration/targets/openssh_keypair/tasks/main.yml
```diff
@@ -4,509 +4,14 @@
...
-    size: 2048
-    passphrase: "{{ passphrase }}"
-    force: true
-  register: privatekey10_result_force
-  when: cryptography_version.stdout is version('3.0', '>=') and bcrypt_version.stdout is version('3.1.5', '>=')
-
-- name: Ensure that ssh-keygen can read keys generated with passphrase
-  command: 'ssh-keygen -yf {{ output_dir }}/privatekey10 -P {{ passphrase }}'
-  register: privatekey10_result_sshkeygen
+- set_fact:
```
felixfontein, line 10: For other modules with multiple backends, there's usually one simple test with backend autodetect before the backend-specific tests, like https://github.com/ansible-collections/community.crypto/blob/main/tests/integration/targets/openssl_privatekey/tasks/main.yml#L34. Basically it only makes sure the module doesn't crash when a backend isn't specified :)

tests/integration/targets/openssh_keypair/tasks/impl.yml
```diff
@@ -0,0 +1,557 @@
...
+  # Ensures no conflicts from previous test runs
+- name: Cleanup Output Directory
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: absent
+  with_fileglob:
+    - "{{ output_dir }}/privatekey*"
+    - "{{ output_dir }}/regenerate*"
+
+- name: Generate privatekey1 - standard (check mode)
```
felixfontein, line 16: For easier debugging, it can be very useful to include the backend's name in very task's name. (Best to use a regexp replace to add it everywhere ;-) )
```suggestion
- name: "({{ backend }}) Generate privatekey1 - standard (check mode)"
```

plugins/module_utils/openssh/backends/keypair_backend.py
```diff
@@ -0,0 +1,476 @@
...
+from ansible_collections.community.crypto.plugins.module_utils.openssh.utils import parse_openssh_version
+from ansible_collections.community.crypto.plugins.module_utils.openssh.cryptography import (
+    HAS_OPENSSH_SUPPORT,
+    HAS_OPENSSH_PRIVATE_FORMAT,
+    InvalidCommentError,
+    InvalidPassphraseError,
+    InvalidPrivateKeyFileError,
+    OpenSSHError,
+    OpensshKeypair,
+
```
felixfontein, line 41:
```suggestion
```

plugins/module_utils/openssh/backends/keypair_backend.py
```diff
@@ -0,0 +1,476 @@
...
+)
+
+
+@six.add_metaclass(abc.ABCMeta)
+class KeypairBackend:
```
felixfontein, line 46:
```suggestion
class KeypairBackend(object):
```
Leaving it away is for Python 3 :)

felixfontein (commented):
Looks good! I only have a few smaller comments :)

felixfontein (approved):
LGTM!

felixfontein:
@Ajpantuso do you want to add more things, or should I just merge this?

Ajpantuso:
> @Ajpantuso do you want to add more things, or should I just merge this?

Nothing more for now. Additional choices for `private_key_format` will be added in a future PR.

felixfontein:
@Ajpantuso thanks a lot for improving this module!

Andersson007:
@Ajpantuso thank you for the great contribution!

PR #239, opened 2021-05-28
[WIP] openssh_cert - add passphrase option for signing key

### LINKED ISSUE #195: Add signing key passphrase parameter to openssh_cert
##### SUMMARY
Allow the user to provide a passphrase to decrypt the signing key when using openssh_cert to create certs.

##### ISSUE TYPE
Not sure if this is a feature or a bug. On the one hand, none of the existing code is broken, per se, so it could be said this is a new feature. On the other hand, this is a pretty severe oversight that makes the module all but useless for anyone who encrypts their private keys, and who doesn't encrypt their private keys, especially a host_ca?

##### COMPONENT NAME
community.crypto.openssh_cert

##### ADDITIONAL INFORMATION

When using community.crypto.openssh_cert to sign a key, and the signing key is encrypted, it is required for the user to provide a passphrase to decrypt it. According to the documentation, community.crypto.openssh_cert has no way to provide this passphrase. When trying to use this module to sign a key using an encrypted ca key, the following error appears:

```
fatal: [23.239.9.7]: FAILED! => {"changed": false, "cmd": "/usr/bin/ssh-keygen -s /path/to/host_ca -h -I '' -n test.example.com -P '' /path/to/ssh_host_rsa_key.pub", "msg": "Load key \"/path/to/host_ca\": incorrect passphrase supplied to decrypt private key", "rc": 255, "stderr": "Load key \"/path/to/host_ca\": incorrect passphrase supplied to decrypt private key\r\n", "stderr_lines": ["Load key \"/path/to/host_ca\": incorrect passphrase supplied to decrypt private key"], "stdout": "", "stdout_lines": []}
```

I found this issue reported on the main ansible repository. It said the issue would be transferred to the community repository, but it doesn't seem to have made its way over here. I am doing that now.

[https://github.com/ansible/ansible/issues/66686](https://github.com/ansible/ansible/issues/66686)

Here is an example of how this could look after it is implemented.

```yaml
- name: Generate an OpenSSH host certificate that is valid forever and only for example.com and examplehost
  community.crypto.openssh_cert:
    type: host
    signing_key: /path/to/private_key
    signing_key_passphrase: "{{ passphrase }}"
    public_key: /path/to/public_key.pub
    path: /path/to/certificate
    valid_from: always
    valid_to: forever
    principals:
        - example.com
        - examplehost

```

Ajpantuso:
##### SUMMARY
Adding `passphrase` option to `openssh_cert` so that an encrypted signing key may be used during certificate generation. `cryptography >= 2.6` is required on the target host to utilize this new feature.

Fixes #195 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
- The encrypted key passed by the user is opened with `cryptography`, decrypted, and written to a temp location already setup in previous versions of this module.
- After `ssh-keygen` is invoked with the decrypted password the temp directory containing any sensitive data is recursively removed.
- No other functional changes are included/intended with this PR.

### DIFF
--- a/plugins/modules/openssh_cert.py
+++ b/plugins/modules/openssh_cert.py
@@ -17,6 +17,7 @@ description:
     - Generate and regenerate OpenSSH host or user certificates.
 requirements:
     - "ssh-keygen"
+    - cryptography >= 2.6 (when I(passphrase) is used)
 options:
     state:
         description:
@@ -46,6 +47,12 @@ options:
             - If the private key is on a PKCS#11 token (I(pkcs11_provider)), set this to the path to the public key instead.
             - Required if I(state) is C(present).
         type: path
+    passphrase:
+        description:
+            - The passphrase used to decrypt the I(signing_key) if the key is encrypted.
+            - Requires cryptography >= 2.6 in addition to ssh-keygen.
+        type: str
+        version_added: 1.8.0
     pkcs11_provider:
         description:
             - To use a signing key that resides on a PKCS#11 token, set this to the name (or full path) of the shared library to use with the token.
@@ -225,11 +232,15 @@ from datetime import MINYEAR, MAXYEAR
 from distutils.version import LooseVersion
 from shutil import copy2, rmtree
 
-from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 from ansible.module_utils._text import to_native
 
 from ansible_collections.community.crypto.plugins.module_utils.crypto.support import convert_relative_to_datetime
 from ansible_collections.community.crypto.plugins.module_utils.openssh.utils import parse_openssh_version
+from ansible_collections.community.crypto.plugins.module_utils.openssh.cryptography import (
+    HAS_OPENSSH_SUPPORT,
+    OpensshKeypair,
+)
 
 
 class CertificateError(Exception):
@@ -247,6 +258,7 @@ class Certificate(object):
         self.pkcs11_provider = module.params['pkcs11_provider']
         self.public_key = module.params['public_key']
         self.path = module.params['path']
+        self.passphrase = module.params['passphrase']
         self.identifier = module.params['identifier']
         self.serial_number = module.params['serial_number']
         self.valid_from = module.params['valid_from']
@@ -274,10 +286,7 @@ class Certificate(object):
     def generate(self, module):
 
         if not self.is_valid(module, perms_required=False) or self.force:
-            args = [
-                self.ssh_keygen,
-                '-s', self.signing_key
-            ]
+            args = [self.ssh_keygen]
 
             if self.pkcs11_provider:
                 args.extend(['-D', self.pkcs11_provider])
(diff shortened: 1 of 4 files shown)


plugins/modules/openssh_cert.py
```diff
@@ -46,6 +47,12 @@
             - If the private key is on a PKCS#11 token (I(pkcs11_provider)), set this to the path to the public key instead.
             - Required if I(state) is C(present).
         type: path
+    passphrase:
```
felixfontein, line 50: How about renaming this to `signing_key_passphrase`? This would be more similar to `privatekey_passphrase` in `openssl_csr`, `x509_certificate` etc.

plugins/modules/openssh_cert.py
```diff
@@ -344,6 +353,21 @@ def generate(self, module):
...
+                # Private key is decrypted and stored in a temporary file when ``passphrase`` value is given
+                if self.passphrase:
+                    keypair = OpensshKeypair.load(
+                        self.signing_key,
+                        passphrase=self.passphrase.encode('UTF-8'),
+                        no_public_key=True
+                    )
+                    keypair.update_passphrase('')
+                    temp_pkey = tempfile.mkstemp(dir=temp_directory)[1]
+                    with open(temp_pkey, 'wb') as f:
```
felixfontein, line 366: Simply writing an unprotected version of the key to disk is somewhat unsatisfactory. Is it possible to pass the password in by stdin instead?

Ajpantuso: So no, OpenSSH added support to load public keys via stdin for fingerprint verificaiton with `-lf '-'`, but otherwise `ssh-keygen` just eats stdin. Any attempts I've made to pass in the data via run_command with/without `prompt_regex` just results in the `Enter passphrase:` prompt halting playbook execution. (I haven't seen OpenSSH acknowledge this behavior anywhere, but that seems to be the consensus from any info I've found)  I don't mind closing this if writing the unencrypted key is non-compliant.

felixfontein: Hmm, I guess it's ok if there's a big fat warning in the option's description that this happens. It's really annoying that OpenSSH doesn't allow to properly script this.

Ajpantuso: I think I'll leave this in draft for a little bit for the following reasons:
- This feature just facilitates automation, but it's nothing that user's wouldn't be able to do themselves (for community.crypto >= 1.7.0)
- In situations where you are permitted to just store CA/signing keys on disk I'm not sure encrypting the keys is a major concern. (i.e. if encryption is really required then you are probably beyond just using `ssh-keygen` with locally stored secrets)
- A more ideal solution would be certificate generation through `cryptography` for which all the mechanisms already exist (See [here](https://github.com/pyca/cryptography/blob/1b922ed1dee0cd7e165a868639ce6d0869c8b2f5/src/cryptography/hazmat/primitives/serialization/ssh.py#L648-L690) for an example where they are already decoding certs) with the exceptions of generating a _nonce_ which can be done with `secrets` and generating a _serial_ which `ssh-keygen` takes as user input anyway. It's just a matter of structuring certificate objects based on type as per the [spec](https://cvsweb.openbsd.org/src/usr.bin/ssh/PROTOCOL.certkeys?annotate=HEAD).

I personally use Vault for ssh key signing so it's not a major priority for me so outside of the 2 people from the original issue on ansible/ansible and the one person here I don't see any urgency to go with a suboptimal hack.

Ajpantuso:
Closing this to prioritize idempotency fixes for this module.

PR #246, opened 2021-06-10
New module utils openssh.certificate

Ajpantuso:
##### SUMMARY
Adding utilities and classes for directly parsing OpenSSH certificates.
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME
plugins/module_utils/openssh/certificate.py

##### ADDITIONAL INFORMATION
Given that `ssh-keygen` does not automate well and the lack of support from common Python libraries implementing SSH I have implemented OpenSSH style certificates in python directly. The original idea was to utilize either [cryptography](https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/primitives/serialization/ssh.py) or [paramiko](https://github.com/paramiko/paramiko/blob/master/paramiko/message.py), but even the backend utilities of these libraries do not fully implement SSH certificates. Starting from where these libraries left off I introduced a "poor man's" reader/parser and writer for SSH data formats as utilities. The primary interface is exposed in `certificate.py` as `OpensshCertificate` which can load relevant certificate information for each supported key type as well as the signing key and signature.

What was left out:
- Certificate generation (This requires `nonce` generation which can be done via `secrets`, but also key signing which would make `cryptography` a dependency)
- Validations for pre-defined `critical_options` or `extensions` (Modules should handle this directly)
- Custom exception classes (Would consider this for the future, but would refactor all exceptions into a single file shared under `module_utils/openssh`


plugins/module_utils/openssh/utils.py
```diff
@@ -33,3 +66,309 @@ def parse_openssh_version(version_string):
...
+        if not isinstance(data, (bytes, bytearray)):
+            raise TypeError("Data must be bytes-like not %s" % type(data))
+
+        self._data = memoryview(data) if PY3 else data
+        self._pos = 0
+
+    def boolean(self):
+        self._check_position(self.BOOLEAN_OFFSET)
+
+        next_pos = self._pos + self.BOOLEAN_OFFSET
```
felixfontein, line 88: How about making `_check_position` return `next_pos`, so you can simply write:
```suggestion
        next_pos = self._check_position(self.BOOLEAN_OFFSET)
```
That would simplify quite a few of the next methods.

plugins/module_utils/openssh/utils.py
```diff
@@ -33,3 +66,309 @@ def parse_openssh_version(version_string):
...
+        self._check_position(offset)
+        self._pos = self._pos + offset
+
+        return self._pos
+
+    def remaining_bytes(self):
+        return len(self._data) - self._pos
+
+    def _check_position(self, offset):
+        if self._pos + offset > len(self._data):
```
felixfontein, line 166: How about first checking that `offset >= 0`? Negative values indicate that something is broken, and also then the `elif` below cannot happen.

felixfontein (commented):
Thanks for your contribution!

I'm not really familiar with the OpenSSH file formats, so I mainly looked at the code itself and not whether it's OpenSSH file handling is correct.

I'm wondering a bit whether it's a good idea to do that much parsing here. It might be better to get this (resp. something like this) placed in cryptography, because a) it gets much more thoroughly checked there, and b) by many more eyes than here.

plugins/module_utils/openssh/certificate.py
```diff
@@ -0,0 +1,337 @@
...
+        self.public_key = public_key
+
+    @property
+    def curve(self):
+        return self._curve
+
+    @curve.setter
+    def curve(self, curve):
+
+        if curve is None:
```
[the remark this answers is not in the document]
Ajpantuso: That was just for initialization so I moved it to __init__()

plugins/module_utils/openssh/utils.py
Ajpantuso: Internally within the parser methods the buffer should only move forward, but `seek` is exposed to manually manipulate the cursor forwards and backwards which is why `_check_position` permits negative offsets.

Ajpantuso:
> Thanks for your contribution!
> 
> I'm not really familiar with the OpenSSH file formats, so I mainly looked at the code itself and not whether it's OpenSSH file handling is correct.
> 
> I'm wondering a bit whether it's a good idea to do that much parsing here. It might be better to get this (resp. something like this) placed in cryptography, because a) it gets much more thoroughly checked there, and b) by many more eyes than here.

I agree cryptography is the better place for this functionality and I can work on implementing it there, but my guess is that would be a much larger discussion and effort.

So my justifications for doing it here first are:
- This does not include the actual cryptographic elements such as generation, signing, and verifying
- This only supplements `ssh-keygen` to facilitate idempotency checks which the output of `ssh-keygen` makes very difficult
- It would avoid a dependency on cryptography which should only be invoked when an interface to `OpenSSL` is required IMO

plugins/module_utils/openssh/utils.py
felixfontein: In that case, all the code locations that use `_check_position` and pass something read by the parser have to make sure that the passed value is non-negative. (Ok, I just noticed that the only place where that happens reads the data with `uint32()`, which is never negative.)

plugins/module_utils/openssh/certificate.py
```diff
@@ -0,0 +1,338 @@
...
+        self.y = parser.mpint()
+
+
+class OpensshECDSACertificateInfo(OpensshCertificateInfo):
+    def __init__(self, curve=None, public_key=None, **kwargs):
+        super(OpensshECDSACertificateInfo, self).__init__(**kwargs)
+        if curve is None:
+            self._curve = curve
+        else:
+            self.curve = curve
```
felixfontein, line 187: How about changing this to:
```suggestion
        self._curve = None
        if curve is not None:
            self.curve = curve
```
This makes it clear that `self._curve` is always present, and if the `curve` setter is ever modified to use the existing `_curve` value, it won't crash ;)

tests/unit/plugins/module_utils/openssh/test_utils.py
```diff
@@ -0,0 +1,178 @@
...
+    assert OpensshParser(OpensshWriter().mpint(mpint).bytes()).mpint() == mpint
+
+
+@pytest.mark.parametrize("mpint", INVALID_MPINT)
+def test_invalid_mpint(mpint):
+    result = False
+
+    try:
+        OpensshWriter().mpint(mpint)
+    except TypeError:
```
felixfontein, line 146: Why not use `with pytest.raises(TypeError):` instead?

felixfontein:
>     * This does not include the actual cryptographic elements such as generation, signing, and verifying
>     * This only supplements `ssh-keygen` to facilitate idempotency checks which the output of `ssh-keygen` makes very difficult

That sounds good.

In that case, I would remove the writer part of the PR, though, to make sure nobody starts abusing it to create / modify objects in the future :)

>     * It would avoid a dependency on cryptography which should only be invoked when an interface to `OpenSSL` is required IMO

In the beginning the module must still work without a dependency on some library that's not widely available on common distros anyway. (It takes years until that's the case.) I think it's still a good long-term goal, and once a cryptography version widely available supports this, I don't see why we can't rely on it. (Hopefully such a version will also support more operations so we can eventually get rid of `ssh-keygen` :) )

plugins/module_utils/openssh/certificate.py
Ajpantuso: > > ```
> > * This does not include the actual cryptographic elements such as generation, signing, and verifying
> > * This only supplements `ssh-keygen` to facilitate idempotency checks which the output of `ssh-keygen` makes very difficult
> > ```
> 
> That sounds good.
> 
> In that case, I would remove the writer part of the PR, though, to make sure nobody starts abusing it to create / modify objects in the future :)
> 
> > ```
> > * It would avoid a dependency on cryptography which should only be invoked when an interface to `OpenSSL` is required IMO
> > ```
> 
> In the beginning the module must still work without a dependency on some library that's not widely available on common distros anyway. (It takes years until that's the case.) I think it's still a good long-term goal, and once a cryptography version widely available supports this, I don't see why we can't rely on it. (Hopefully such a version will also support more operations so we can eventually get rid of `ssh-keygen` :) )

I do need several of the writer's functions to determine public key fingerprints:

```python
    def public_key_fingerprint(self):
        if any([self.e is None, self.n is None]):
            return b''

        writer = OpensshWriter()
        writer.string(_SSH_TYPE_STRINGS['rsa'])
        writer.mpint(self.e)
        writer.mpint(self.n)

        return fingerprint(writer.bytes())
```
I put a warning in the docstring if that's acceptable.
Otherwise I can move the required functions into `certificate.py` and name them to be unexported.

Ajpantuso:
> > ```
> > * This does not include the actual cryptographic elements such as generation, signing, and verifying
> > * This only supplements `ssh-keygen` to facilitate idempotency checks which the output of `ssh-keygen` makes very difficult
> > ```
> 
> That sounds good.
> 
> In that case, I would remove the writer part of the PR, though, to make sure nobody starts abusing it to create / modify objects in the future :)
> 
> > ```
> > * It would avoid a dependency on cryptography which should only be invoked when an interface to `OpenSSL` is required IMO
> > ```
> 
> In the beginning the module must still work without a dependency on some library that's not widely available on common distros anyway. (It takes years until that's the case.) I think it's still a good long-term goal, and once a cryptography version widely available supports this, I don't see why we can't rely on it. (Hopefully such a version will also support more operations so we can eventually get rid of `ssh-keygen` :) )

I do need several of the writer's functions to determine public key fingerprints:

```python
    def public_key_fingerprint(self):
        if any([self.e is None, self.n is None]):
            return b''

        writer = OpensshWriter()
        writer.string(_SSH_TYPE_STRINGS['rsa'])
        writer.mpint(self.e)
        writer.mpint(self.n)

        return fingerprint(writer.bytes())
```
I put a warning in the docstring if that's acceptable.
Otherwise I can move the required functions into `certificate.py` and name them to be unexported.

tests/unit/plugins/module_utils/openssh/test_utils.py
Ajpantuso: Done

plugins/module_utils/openssh/certificate.py
Ajpantuso: Makes sense, added in the last commit.

felixfontein:
> I do need several of the writer's functions to determine public key fingerprints:

Ah, I somehow missed these when grepping.

> I put a warning in the docstring if that's acceptable.
> Otherwise I can move the required functions into `certificate.py` and name them to be unexported.

How about keeping them in that file, but making them internal by renaming `OpensshWriter` to `_OpensshWriter` and mentioning in the docstring that this is a private API that must not be used except by the openssh module_utils?

Ajpantuso:
> How about keeping them in that file, but making them internal by renaming `OpensshWriter` to `_OpensshWriter` and mentioning in the docstring that this is a private API that must not be used except by the openssh module_utils?

Done, thanks.

felixfontein (approved):

felixfontein:
@Ajpantuso thanks a lot for this contribution!

PR #255, opened 2021-07-14
openssh_cert - cleanup and diff support

Ajpantuso:
##### SUMMARY
Significant cleanup of `openssh_cert` with added `diff` support;

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

#### Changes:
- Added `diff` support after integration of `OpensshCertificate` from module_utils
- Refactored integration tests and enabled assertions so that idempotency is always checked. (Also added test cases for known idempotency issues to be addressed in a future PR)
- Time parameters presented in the "validity string" for the `-V` option will now just use `always` and `forever` for arbitrarily large and small dates rather than `datetime.min` and `datetime.max` from `Python`. (No effect to end users as `ssh-keygen` will behave the same way as when provided out-of-range dates)

#### Fixes:
- Used `AnsibleModule` methods to more consistently setup temp files and preserve permissions.
- Ensured backup is taken before final move operation to restore if necessary.


plugins/module_utils/openssh/certificate.py
```diff
@@ -69,6 +71,107 @@
...
+        return result
+
+    def to_datetime(self, time_string_or_timestamp):
+        try:
+            if isinstance(time_string_or_timestamp, str):
+                result = self._time_string_to_datetime(time_string_or_timestamp.strip())
+            elif isinstance(time_string_or_timestamp, (long, int)):
+                result = self._timestamp_to_datetime(time_string_or_timestamp)
+            else:
+                raise ValueError
```
felixfontein, line 126:
```suggestion
                raise ValueError('Parameter needs to be string or integer')
```

plugins/module_utils/openssh/certificate.py
```diff
@@ -69,6 +71,107 @@
...
+            else:
+                result = dt.isoformat() if date_format == 'human_readable' else dt.strftime("%Y%m%d%H%M%S")
+        elif date_format == 'timestamp':
+            td = dt - _ALWAYS
+            result = int((td.microseconds + (td.seconds + td.days * 24 * 3600) * 10 ** 6) / 10 ** 6)
+        else:
+            raise ValueError("%s is not a valid format" % date_format)
+        return result
+
+    def to_datetime(self, time_string_or_timestamp):
```
felixfontein, line 119: Shouldn't this be a @staticmethod as well, since it only uses static methods?

felixfontein (commented):
Looks good, thanks a lot for working on this!

felixfontein (approved):

felixfontein:
@Ajpantuso thanks for implementing this!

Ajpantuso:
> @Ajpantuso thanks for implementing this!

Thanks for reviewing. The follow-up PR's should be much smaller.

Issue #20, opened 2020-04-05
openssh_cert: idempotency check ignores signing_key, public_key, options and identifier


felixfontein:
##### SUMMARY
The idempotency check of `openssh_cert` ignores changes to `signing_key`, `public_key`, `options` and `identifier`.

Reported by @das7pad [here](https://github.com/ansible/ansible/pull/54653#issuecomment-478339894).

(Migration of ansible/ansible#54655)

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
openssh_cert

##### ANSIBLE VERSION
```paste below
2.8.0
```

felixfontein:
Closing since this has been mitigated in #256.

PR #256, opened 2021-07-16
openssh_cert - Adding regenerate option

Ajpantuso:
##### SUMMARY
Adds `regenerate` option to `openssh_cert` so that `identifier`, `options`, `public_key`, and `signing_key` are also checked for idempotency.
Implements #20 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
##### ADDITIONAL INFORMATION
- Defaults to pre 1.8.0 behavior to avoid a `breaking_change`
- Minor changes to interfaces were made as well, but only to code releasing with `1.8.0`


plugins/modules/openssh_cert.py
```diff
@@ -36,6 +36,18 @@
...
+    idempotency:
+        description:
+            - When C(partial) and I(force=false) an existing certificate will be regenerated based on
+              I(serial), I(type), I(valid_from), I(valid_to), I(valid_at), and I(principals).
+            - When C(full) and I(force=false), I(identifier), I(options), I(public_key), and I(signing_key)
+              are also considered when compared against an existing certificate.
+        type: str
+        choices:
+            - partial
+            - full
```
felixfontein, line 48: How about adding an `always_regenerate` option, and document `force` as "will set `idempotency` to `always_regenerate`"?

But maybe it's better to rename this option to `regenerate` so it can be more similar to the option of the same name of openssl_privatekey. Have you seen that option?

Ajpantuso: Yes, I had to work with that pattern for `openssh_keypair` which has a similar implementation. Adding `never` and `always` would be trivial so I can implement the same here.

plugins/module_utils/openssh/certificate.py
```diff
@@ -174,6 +203,78 @@ def _time_string_to_datetime(time_string):
         return result
 
 
+class OpensshCertificateOption(object):
+    def __init__(self, option_type, name, data):
+        if not isinstance(option_type, six.string_types) or option_type not in ('critical', 'extension'):
```
felixfontein, line 208: This condition is stricter than the other:
```suggestion
        if option_type not in ('critical', 'extension'):
```

plugins/modules/openssh_cert.py
```diff
@@ -34,13 +34,34 @@
...
         default: false
     path:
         description:
             - Path of the file containing the certificate.
         type: path
         required: true
+    regenerate:
+        description:
+            - When C(never) the task will fail if a certificate already exists at I(path) and is unreadable
+              otherwise a new certificate will only be generated if there is no existing certificate.
```
felixfontein, line 48:
```suggestion
            - When C(never) the task will fail if a certificate already exists at I(path) and is unreadable.
              Otherwise a new certificate will only be generated if there is no existing certificate.
```

plugins/modules/openssh_cert.py
```diff
@@ -377,34 +394,83 @@ def _command_arguments(self, key_copy_path):
...
-                    valid_from=self.original_data.valid_after,
-                    valid_to=self.original_data.valid_before
+        partial_result = all([
+            set(self.original_data.principals) == set(self.principals),
+            self.original_data.serial == self.serial_number if self.serial_number is not None else True,
+            self.original_data.type == self.type,
+            self._compare_time_parameters(),
+        ])
+
+        return partial_result if self.regenerate == 'partial_idempotence' else partial_result and all([
```
felixfontein, line 453: I think this way it would be a bit easier to understand:
```suggestion
        if self.regenerate == 'partial_idempotence':
            return partial_result

        return partial_result and all([
```

felixfontein (approved):
Looks good, thanks!

(Sorry for the delay, I forgot looking back here during/after vacation...)

PR #257, opened 2021-07-20
fix custom file attributes for public keys

crwr45:
Use of the confusingly-named _permissions_changed() on both
sides of an `or` was resulting in the second invocation not
being reached if the first invocation returned True.

This is fixed by explicitly running both invocations
before evaluating the return values.

Signed-off-by: Charlie Wheeler-Robinson <cwheeler@redhat.com>

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/changelogs/fragments/257-openssh-keypair-fix-pubkey-permissions.yml
+++ b/changelogs/fragments/257-openssh-keypair-fix-pubkey-permissions.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - openssh_keypair - fixed a bug that prevented custom file attributes being applied to public keys (https://github.com/ansible-collections/community.crypto/pull/257).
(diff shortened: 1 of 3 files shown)


felixfontein:
Thanks for your contributoin! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

Ajpantuso (approved):
LGTM, thanks for the fix!
The generation and validation logic in this module is still pending re-write and one of the items is to make `_permissions_changed` into a command like `update_permissions` which updates `self.changed` directly.

felixfontein (approved):

felixfontein:
@crwr45 thanks for fixing this!

PR #259, opened 2021-08-04
openssh_keypair - integration test refactoring

Ajpantuso:
##### SUMMARY
Breaks up `openssh_keypair` integration tests in preparation for a follow-up PR which cleans up the module code.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
tests/integration/targets/openssh_keypair/tasks/main.yml

##### ADDITIONAL INFORMATION
- There's still some overlap between `regenerate` and the other test files, but safer to leave those for now.
- Adds some checks that were missing for file options such as `mode`.


felixfontein:
@Ajpantuso it seems that some tests are still failing.

BTW, the next Ansible release is next week, so if we want to get the [current changes](https://github.com/ansible-collections/community.crypto/tree/main/changelogs/fragments) (mostly your changes :) ) in it, we should release 1.8.0 before next Tuesday.

Ajpantuso:
> @Ajpantuso it seems that some tests are still failing.
> 
> BTW, the next Ansible release is next week, so if we want to get the [current changes](https://github.com/ansible-collections/community.crypto/tree/main/changelogs/fragments) (mostly your changes :) ) in it, we should release 1.8.0 before next Tuesday.

Sorry, tests should be good now.

I'm fine with releasing 1.8.0 without this PR and the [follow-up](https://github.com/Ajpantuso/community.crypto/commit/13589e9bd5db4dc2292781bc63565a2a915926d7). The burden would be to review these in time and the second PR is substantial. From a user perspective it just adds `diff` support  and proper handling of backups to `openssh_keypair`, but internally it brings cohesion to the `openssh` module_utils.

tests/integration/targets/openssh_keypair/tasks/main.yml
```diff
@@ -17,7 +17,30 @@
     backends: "{{ backends + ['cryptography'] }}"
   when: cryptography_version.stdout is version('3.0', '>=') and bcrypt_version.stdout is version('3.1.5', '>=')
 
-- include_tasks: ./impl.yml
+- include_tasks: ../tests/core.yml
```
felixfontein, line 20: Out of curiosity: why not have all files in `tasks/`?

felixfontein:
(The last test already passed; it seems that a webhook got lost and it's just GH that doesn't know...)

felixfontein (approved):

felixfontein:
@Ajpantuso thanks for improving the tests!

tests/integration/targets/openssh_keypair/tasks/main.yml
Ajpantuso: No good reason outside of it matching what I did with [openssh_cert](https://github.com/ansible-collections/community.crypto/tree/main/tests/integration/targets/openssh_cert/tests).

PR #260, opened 2021-08-04
openssh_keypair - Add diff support and general cleanup

Ajpantuso:
##### SUMMARY
Adds `diff` support for `openssh_keypair` as well as enhances file overwriting operations to preserve and restore from backup in the event of failure. Also integrates both of the `openssh_*` modules to ensure consistency.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/openssh_keypair.py
##### ADDITIONAL INFORMATION
Fixes some issues for file permissions as well:
- openssh_keypair - `ssh-keygen` preserves the original files permissions when overwriting and these changes ensure the `cryptography` backend does the same.
- both modules - Ensures that `atomic_move` is replaced with `preserved_copy` when no file exists at the destination to prevent a potentially secure file from being updated with permissions from the parent directory's mask.

### DIFF
--- a/plugins/module_utils/openssh/backends/common.py
+++ b/plugins/module_utils/openssh/backends/common.py
@@ -31,19 +31,17 @@ from ansible_collections.community.crypto.plugins.module_utils.openssh.utils imp
 
 def _restore_on_failure(f):
     def backup_and_restore(self, sources_and_destinations, *args, **kwargs):
-        backups = [(d, self.module.backup_local(d) if os.path.exists(d) else None) for s, d in sources_and_destinations]
+        backups = [(d, self.module.backup_local(d)) for s, d in sources_and_destinations if os.path.exists(d)]
 
         try:
             f(self, sources_and_destinations, *args, **kwargs)
         except Exception:
             for destination, backup in backups:
-                if backup is not None:
-                    self.module.atomic_move(backup, destination)
+                self.module.atomic_move(backup, destination)
             raise
         else:
             for destination, backup in backups:
-                if backup is not None:
-                    self.module.add_cleanup_file(backup)
+                self.module.add_cleanup_file(backup)
     return backup_and_restore
 
 
@@ -55,15 +53,9 @@ class OpensshModule(object):
         self.changed = False
         self.check_mode = self.module.check_mode
 
-    @property
-    @abc.abstractmethod
-    def diff(self):
-        pass
-
-    @property
-    @abc.abstractmethod
-    def _result(self):
-        pass
+    def execute(self):
+        self._execute()
+        self.module.exit_json(**self.result)
 
     @abc.abstractmethod
     def _execute(self):
@@ -80,9 +72,15 @@ class OpensshModule(object):
 
         return result
 
-    def execute(self):
-        self._execute()
-        self.module.exit_json(**self.result)
+    @property
+    @abc.abstractmethod
+    def _result(self):
+        pass
+
+    @property
+    @abc.abstractmethod
+    def diff(self):
+        pass
 
     @staticmethod
     def skip_if_check_mode(f):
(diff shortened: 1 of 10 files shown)


felixfontein:
Please note that ansible-test's recently updated pylint reports some failures for this PR.

Ajpantuso:
> Please note that ansible-test's recently updated pylint reports some failures for this PR.

Addressed with the latest commit.

plugins/module_utils/openssh/backends/common.py
```diff
@@ -18,25 +18,289 @@
...
-def restore_on_failure(f):
-    def backup_and_restore(module, path, *args, **kwargs):
-        backup_file = module.backup_local(path) if os.path.exists(path) else None
+from ansible_collections.community.crypto.plugins.module_utils.openssh.utils import (
+    parse_openssh_version,
+)
+
+
+def _restore_on_failure(f):
+    def backup_and_restore(self, sources_and_destinations, *args, **kwargs):
```
felixfontein, line 33: `restore_on_failure` is part of the public API of this collection (even though probably nobody uses it), so these changes are breaking ones.

How about keeping `restore_on_failure` without underscore, and adding something like
```
if not isinstance(sources_and_destinations, (list, tuple)):
    sources_and_destinations = [(sources_and_destinations, sources_and_destinations)]
```

plugins/module_utils/openssh/backends/common.py
```diff
@@ -18,25 +18,289 @@
...
+        backups = [(d, self.module.backup_local(d) if os.path.exists(d) else None) for s, d in sources_and_destinations]
```
felixfontein, line 34:
```suggestion
        backups = [(d, self.module.backup_local(d)) for s, d in sources_and_destinations if os.path.exists(d)]
```

plugins/module_utils/openssh/backends/common.py
```diff
@@ -18,25 +18,289 @@
...
 
         try:
-            f(module, path, *args, **kwargs)
+            f(self, sources_and_destinations, *args, **kwargs)
         except Exception:
-            if backup_file is not None:
-                module.atomic_move(backup_file, path)
+            for destination, backup in backups:
+                if backup is not None:
```
felixfontein, line 40: With the above change, this (and the similar line below) wouldn't be needed anymore.

plugins/module_utils/openssh/backends/common.py
```diff
@@ -18,25 +18,289 @@
...
-
+            for destination, backup in backups:
+                if backup is not None:
+                    self.module.add_cleanup_file(backup)
     return backup_and_restore
 
 
-@restore_on_failure
-def safe_atomic_move(module, path, destination):
-    module.atomic_move(path, destination)
```
felixfontein, line 42: This function needs to stay as well.

felixfontein (commented):
Looks good! I mainly have a few comments on some helpers.

Ajpantuso:
> Looks good! I mainly have a few comments on some helpers.

Done, I had also moved around some method definitions to make the classes more readable, but no functional changes in the last two commits.

plugins/module_utils/openssh/backends/common.py
felixfontein: Hmmm, one thing I missed is that previously, the first parameter was `module`, while now the module is `self.module`. I'm not sure what's the best way to fix this...

Ajpantuso: I'll just revert `restore_on_failure` and create a new decorator for the desired functionality.

felixfontein (approved):

felixfontein:
@Ajpantuso thanks a lot for implementing this!

PR #267, opened 2021-09-01
Make ready for split-controller testing in ansible-core

felixfontein:
##### SUMMARY
Uses ansible/ansible#75605 instead of devel in CI. Let's see what happens and if there are surprises :-)

##### ISSUE TYPE
- CI Pull Request

##### COMPONENT NAME
CI

PR #268, opened 2021-09-07
Improve CI

felixfontein:
##### SUMMARY
Contains all fixes from #267 that are not specific to split-controller testing. These fixes weren't needed so far since for ansible-test, controller == remote. This will change with split-controller testing, and thus these fixes are necessary.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
integration tests

### DIFF
--- a/tests/integration/targets/acme_account/meta/main.yml
+++ b/tests/integration/targets/acme_account/meta/main.yml
@@ -1,2 +1,3 @@
 dependencies:
   - setup_acme
+  - setup_remote_tmp_dir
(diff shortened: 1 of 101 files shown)


briantist (approved):
despite being a large boi, this diff was a breeze to review, very straightforward in my opinion, nice work!

felixfontein:
Since this are mainly mechanical changes to the tests, and they still pass, I'm pretty sure they don't break something. So merging.

PR #269, opened 2021-09-09
openssl_csr: fix error in docs

felixfontein:
##### SUMMARY
As @noonedeadpunk pointed out in IRC, it's authority_cert_issuer and authority_cert_serial_number that must be specified together (as also the code confirms, and the cryptography library code as well).

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
openssl_csr

### DIFF
--- a/plugins/doc_fragments/module_csr.py
+++ b/plugins/doc_fragments/module_csr.py
@@ -231,7 +231,7 @@ options:
                own choice. Specifying this option is mostly useful for self-signed certificates
                or for own CAs."
             - Note that this is only supported if the C(cryptography) backend is used!
-            - The C(AuthorityKeyIdentifier) will only be added if at least one of I(authority_key_identifier),
+            - The C(AuthorityKeyIdentifier) extension will only be added if at least one of I(authority_key_identifier),
               I(authority_cert_issuer) and I(authority_cert_serial_number) is specified.
         type: str
     authority_cert_issuer:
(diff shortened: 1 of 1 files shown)


noonedeadpunk:
LGTM

Ajpantuso (approved):
LGTM, but one comment:
The following phrase appears for each of the affected options (lines 234-235, 248-249, 260-261 of this PR)
```yaml
- The C(AuthorityKeyIdentifier) will only be added if at least one of I(authority_key_identifier),
  I(authority_cert_issuer) and I(authority_cert_serial_number) is specified.
```
however the first part of the phrase seems self-evident and the second part of the phrase only really needs to be mentioned for `authority_key_identifier`.

felixfontein:
> ```yaml
> - The C(AuthorityKeyIdentifier) will only be added if at least one of I(authority_key_identifier),
>   I(authority_cert_issuer) and I(authority_cert_serial_number) is specified.
> ```

This is missing *extension* after `C(...)`; I've added that in 6e618c4b10fb39b388371d933dbf6e64b4138087.

> however the first part of the phrase seems self-evident and the second part of the phrase only really needs to be mentioned for `authority_key_identifier`.

I'm not sure whether it's really fully self-evident. I think it's better to be explicit and mention when exactly that extension will be added (namely when at least one of these three options is specified).

Ajpantuso:
Adding "extension" does make it more clear why this statement is needed (and for all three options). Thanks for clarifying!

felixfontein:
Thanks for bringing this up, I had to read it a couple of times to notice that it was actually missing a word ;-) (From the remainder of the sentence, I'm pretty sure I intended it to be in there when I wrote it...)

felixfontein:
@noonedeadpunk @Ajpantuso thanks for reviewing this!

PR #274, opened 2021-09-12
Make Dirname (de)serialization conformant to RFC 4514

### LINKED ISSUE #101: Make dirName parsing / outputting more conformant to RFC4514
##### SUMMARY
See https://tools.ietf.org/html/rfc4514.html#section-3 (from string) and https://tools.ietf.org/html/rfc4514.html#section-2 (to string). Also see https://tools.ietf.org/html/rfc4514.html#section-2.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
plugins/module_utils/crypto/cryptography_support.py

felixfontein:
##### SUMMARY
Fixes #101.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/module_utils/crypto/cryptography_support.py

### DIFF
--- a/tests/unit/plugins/module_utils/crypto/test_cryptography_support.py
+++ b/tests/unit/plugins/module_utils/crypto/test_cryptography_support.py
@@ -44,16 +44,21 @@ def test_cryptography_get_name_other_name_utfstring():
 
 
 @pytest.mark.parametrize('name, options, expected', [
-    (u'CN=x ', {}, (NameAttribute(oid.NameOID.COMMON_NAME, u'x '), u'')),
-    (u'CN=\\ ', {}, (NameAttribute(oid.NameOID.COMMON_NAME, u' '), u'')),
-    (u'CN=\\#', {}, (NameAttribute(oid.NameOID.COMMON_NAME, u'#'), u'')),
-    (u'CN=#402032', {}, (NameAttribute(oid.NameOID.COMMON_NAME, u'@ 2'), u'')),
-    (u'CN = x ', {}, (NameAttribute(oid.NameOID.COMMON_NAME, u'x '), u'')),
-    (u'CN = x\\, ', {}, (NameAttribute(oid.NameOID.COMMON_NAME, u'x, '), u'')),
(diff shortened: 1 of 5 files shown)


tests/unit/plugins/module_utils/crypto/test_cryptography_support.py
```diff
@@ -31,3 +41,64 @@ def test_cryptography_get_name_other_name_utfstring():
...
+def test_parse_dn_component_failure(name, options, message):
+    with pytest.raises(OpenSSLObjectError, match=u'^%s$' % re.escape(message)):
+        result = _parse_dn_component(name, **options)
+
+
+@pytest.mark.parametrize('name, expected', [
+    (u'CN=foo', [NameAttribute(oid.NameOID.COMMON_NAME, u'foo')]),
+    (u'CN=foo,CN=bar', [NameAttribute(oid.NameOID.COMMON_NAME, u'foo'), NameAttribute(oid.NameOID.COMMON_NAME, u'bar')]),
+    (u'CN  =  foo ,  CN  =  bar', [NameAttribute(oid.NameOID.COMMON_NAME, u'foo '), NameAttribute(oid.NameOID.COMMON_NAME, u'bar')]),
+])
```
Ajpantuso, line 91: Everything looks good to me. I think one non-ascii test-case, as in the [RFC examples](https://datatracker.ietf.org/doc/html/rfc4514#section-4), would be nice (really just a sanity check that non-ascii UTF-8 characters are not escaped).

felixfontein: That was a really good idea, since it pointed out one flaw in the existing code: working with unicode strings doesn't allow to escape unicode endpoints with 8-bit hex digits. I've rewritten it to use byte strings, and added some examples from the RFC as tests. Looks much better now IMO.

Ajpantuso (approved):
LGTM

tests/unit/plugins/module_utils/crypto/test_cryptography_support.py
Ajpantuso: Nice!

felixfontein:
Restarting CI.

felixfontein:

PR #281, opened 2021-09-15
Improve CI

felixfontein:
##### SUMMARY
CI improvements from #267.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
integration tests


felixfontein:
ready for review

tests/integration/targets/acme_certificate/tasks/impl.yml
```diff
@@ -264,92 +264,98 @@
...
+        # All intermediates have the same subject key identifier, so always
+        # the first chain will be found, and we need a second condition to
+        # make sure that the first condition actually works. (The second
+        # condition has been tested above.)
+        - test_certificates: first
+          subject_key_identifier: "{{ acme_intermediates[0].subject_key_identifier }}"
+        - test_certificates: last
+          issuer: "{{ acme_roots[1].subject }}"
+      use_csr_content: true
+  - name: Store obtain results for cert 6
```
Andersson007, line 296:
```suggestion
      use_csr_content: true

    - name: Store obtain results for cert 6
```
I suspect that this is a common style for the tests in this collection, isn't it? IMO would be nice to split a bit

felixfontein: Unfortunately it is, at least since the collection was separated from ansible/ansible. I think the migration script removed all empty lines in tests... It would definitely be nice to split a bit, though I think that's better done in a separate PR.

felixfontein:
@Andersson007 thanks for reviewing this!

tests/integration/targets/acme_certificate/tasks/impl.yml
Andersson007: makes sense!

Andersson007:
@felixfontein thanks for improving CI!

PR #282, opened 2021-09-17
acme_challenge_cert_helper: fail better to avoid crashes in Ansible

felixfontein:
##### SUMMARY
Modules which return `exception=None` in fail_json risk the Ansible callback to crash when trying to process the exception. Prevent this by only returning `exception` if it is not `None`. (It can be `None` if cryptography was found, but a version before 1.3.)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
acme_challenge_cert_helper


Ajpantuso (approved):
LGTM, thanks!

PR #287, opened 2021-09-28
Remove vendored copy of ipaddress

felixfontein:
##### SUMMARY
For 2.0.0, we should remove the vendored ipaddress copy. It's included in the stdlib of Python 3, and is also a dependency of older cryptography versions (which did not require Python 3).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ipaddress compat module_utils


felixfontein:
ready_for_review

plugins/module_utils/acme/acme.py
```diff
@@ -38,6 +39,14 @@
     nopad_b64,
 )
 
+try:
+    import ipaddress
+except ImportError:
+    HAS_IPADDRESS = False
+    IPADDRESS_IMPORT_ERROR = traceback.format_exc()
+else:
+    HAS_IPADDRESS = True
```
webknjaz, line 48: `traceback.format_exc()` is an overkill. Just do
```suggestion
try:
    import ipaddress
except ImportError as import_exc:
    IPADDRESS_IMPORT_ERROR = import_exc
else:
    IPADDRESS_IMPORT_ERROR = None
    del ipaddress  # prevent accidental use later in the code
```
Also, `HAS_IPADDRESS` is not really necessary as you can reuse the exception var to signal that it's missing.

plugins/module_utils/acme/acme.py
```diff
@@ -12,6 +12,7 @@
 import datetime
 import json
 import locale
+import traceback
```
webknjaz, line 15:
```suggestion
```

plugins/module_utils/acme/acme.py
```diff
@@ -327,6 +336,9 @@ def get_default_argspec():
 
 
 def create_backend(module, needs_acme_v2):
+    if not HAS_IPADDRESS:
+        module.fail_json(msg=missing_required_lib('ipaddress'), exception=IPADDRESS_IMPORT_ERROR)
```
webknjaz, line 340:
```suggestion
    if IPADDRESS_IMPORT_ERROR is not None:
        module.fail_json(msg=missing_required_lib('ipaddress'), exception=IPADDRESS_IMPORT_ERROR)
```

plugins/module_utils/acme/challenges.py
```diff
@@ -13,11 +13,10 @@
 import json
 import re
 import time
+import traceback
```
webknjaz, line 16:
```suggestion
```
(unused import)

plugins/module_utils/acme/challenges.py
```diff
@@ -28,6 +27,11 @@
     ModuleFailException,
 )
 
+try:
+    import ipaddress
+except ImportError:
+    pass
```
webknjaz, line 33: Is this really necessary if it's shielded against being missing on the outer layer?
```suggestion
import ipaddress
```

tests/unit/plugins/module_utils/acme/test_backend_openssl_cli.py
```diff
@@ -24,7 +24,6 @@
     ("0000:0001:0000:0000:0001:0000:0000:0001", "0:1::1:0:0:1"),
     ("0000:0001:0000:0001:0000:0001:0000:0001", "0:1:0:1:0:1:0:1"),
     ("0.0.0.0", "0.0.0.0"),
-    ("000.001.000.000", "0.1.0.0"),
```
webknjaz, line 27: reason?

plugins/module_utils/acme/backend_openssl_cli.py
```diff
@@ -29,7 +29,10 @@
 
 from ansible_collections.community.crypto.plugins.module_utils.acme.utils import nopad_b64
 
-from ansible_collections.community.crypto.plugins.module_utils.compat import ipaddress as compat_ipaddress
+try:
+    import ipaddress
+except ImportError:
+    pass
```
webknjaz, line 35: same
```suggestion
import ipaddress
```

webknjaz (commented):
Here's how you could simplify this a little.

plugins/module_utils/acme/acme.py
tadeboro: Since the code Felix used follows the https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_best_practices.html#importing-and-using-shared-code, I would say that if we need a new way of doing things, we should first update the official docs.

tests/unit/plugins/module_utils/acme/test_backend_openssl_cli.py
felixfontein: Such IPs are no longer accepted by ipaddress in Python stdlib (see CVE-2021-29921, https://sick.codes/sick-2021-014/).

plugins/module_utils/acme/challenges.py
felixfontein: Yes. Also otherwise the sanity import tests will fail, since without this, this file is no longer importable on a Python 2.6 or 2.7 without ipaddress installed.

tests/unit/plugins/module_utils/acme/test_backend_openssl_cli.py
webknjaz: Ah, it'd be good to mention in the commit message or somewhere.

plugins/module_utils/acme/acme.py
webknjaz: Makes sense. It's just that this is not applicable to every use-case.

plugins/module_utils/acme/acme.py
felixfontein: Does `fail_json` (and `exit_json`) accept exception objects? At least the module_utils in ansible-core still actively use `traceback.format_exc()` instead of just passing the exception objects to `fail_json`.

plugins/module_utils/acme/challenges.py
webknjaz: Also, here's a way to do the same that better reflects the semantics of the intent to ignore the importing outcome. It reads as "attempt to make an import but suppress a failure to do so".
```python
import contextlib
```
```suggestion
with contextlib.suppress(ImportError):
    import ipaddress
```
(of course, if the Python runtime has this, I don't remember when this was added)

tests/unit/plugins/module_utils/acme/test_backend_openssl_cli.py
felixfontein: 730e9e580af8805e1f3b59b1d896fa5a1ace11cc now mentions it.

felixfontein:
@tadeboro @webknjaz thanks for reviewing this!

PR #294, opened 2021-09-30
Support cryptography 35.0.0 for all modules except openssl_pkcs12

felixfontein:
##### SUMMARY
The latest cryptography release, 35.0.0 (released today), uses a lot more Rust code than previous relesaes. Unfortunately this breaks some of our 'workarounds' which access some internals to gather some information that cryptography unfortunately does not provide. These are in particular:

1. Getting hold of raw extension data (DER encoded bytes) for all CSR and X.509 certificate extensions (needed for the openssl_csr_info, x509_certificate_info, and get_certificate modules);
2. Getting hold of the 'friendly name' of the 'main' certificate in a PKCS#12 file (needed for the openssl_pkcs12 module).

I've been able to work around 1 in some ugly fashion, but didn't find a working solution for 2 yet. It seems that the friendly name is somehow stripped from the certificate.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cryptography support

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/crypto/cryptography_support.py
+++ b/plugins/module_utils/crypto/cryptography_support.py
@@ -30,6 +30,7 @@ from ._asn1 import serialize_asn1_string_as_der
 try:
     import cryptography
     from cryptography import x509
+    from cryptography.hazmat.backends import default_backend
     from cryptography.hazmat.primitives import serialization
     import ipaddress
 except ImportError:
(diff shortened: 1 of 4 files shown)


plugins/module_utils/crypto/cryptography_support.py
```diff
@@ -68,24 +69,53 @@ def cryptography_get_extensions_from_cert(cert):
     # Since cryptography won't give us the DER value for an extension
     # (that is only stored for unrecognized extensions), we have to re-do
     # the extension parsing outselves.
+    backend = default_backend()
+    try:
+        # For certain old versions of cryptography, backend is a MultiBackend object,
+        # which has no _lib attribute. In that case, revert to the old approach.
+        x = backend._lib
```
webknjaz, line 76: Did you mean 
```suggestion
        backend._lib
```

plugins/module_utils/crypto/cryptography_support.py
```diff
@@ -68,24 +69,53 @@ def cryptography_get_extensions_from_cert(cert):
...
-        entry = dict(
-            critical=(crit == 1),
-            value=base64.b64encode(der),
-        )
-        oid = obj2txt(backend._lib, backend._ffi, backend._lib.X509_EXTENSION_get_object(ext))
-        result[oid] = entry
+    try:
+        x509_obj = cert._x509
+
+        for i in range(backend._lib.X509_get_ext_count(x509_obj)):
```
webknjaz, line 84: It is a code smell to wrap big blocks of code with try-except. If anything unexpected on any line raises `AttributeError`, expect surprises in the control flow. You could probably put this in the else-block, though.

plugins/module_utils/crypto/cryptography_support.py
felixfontein: Both should have the same result. But yeah, removing that assignment is probably a good idea. I did it in all places at once in 5a27b231214e318156d79220df3ea89347879377.

plugins/module_utils/crypto/cryptography_support.py
felixfontein: I rewrote it to minimize the amount of lines covered by `try`. This is somewhat less efficient, since the decision which code-path is taken is now done once per iteration, but it definitely improves readability. I hope that sooner or later we can get rid of these functions anyway, resp. rewrite them so that the happy path doesn't require any internals.

Changed in 548cdc698579d9d47d44bef626eeda47b05da039.

felixfontein:
The follow-up #296 will fix the remaining PKCS#12 issues (once rebased after this has been merged).

briantist (approved):
unfortunate to have to resort to such workarounds, but these look like solid fixes to me

felixfontein:
@briantist I'm really looking forward to get rid of them (resp. only have to use them for older cryptography versions)! I've started another push for that in cryptography, but it probably has to wait a bit longer (https://github.com/pyca/cryptography/pull/6346#issuecomment-932606291)... Let's see how many ugly workarounds similar to the ones here or in #296 we need until then...

felixfontein:
@webknjaz @briantist thanks a lot for reviewing this!

Issue #295, opened 2021-09-30
x509_certificate does not recreate certificate if ownca_not_after was changed


kbudde:
##### SUMMARY
Certificate is not recreated after `ownca_not_after`/`selfsigned_not_after` was changed.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.crypto.x509_certificate

##### ANSIBLE VERSION
```
ansible 2.10.5
  config file = None
  configured module search path = ['/home/kris/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.5 (default, May 11 2021, 08:20:37) [GCC 10.3.0]
```

##### COLLECTION VERSION
```
# /usr/lib/python3/dist-packages/ansible_collections
Collection       Version
---------------- -------
community.crypto 1.4.0  

# /home/kris/.ansible/collections/ansible_collections
Collection       Version
---------------- -------
community.crypto 1.9.4
```

##### CONFIGURATION
fresh install / nothing changed

##### OS / ENVIRONMENT
Ubuntu 20.04

##### STEPS TO REPRODUCE
I created a CA and a server certificate.
I've updated the validity (`ownca_not_after`) but the certificate was not changed. Same behavior with `selfsigned_not_after` on ca creation.

I created a playbook to demonstrate the behavior.

```yaml
  - name: Create & sign the server cert with 30days validity
    community.crypto.x509_certificate:
      path: server.crt
      csr_path: server.csr
      provider: ownca
      ownca_path: ca.crt
      ownca_privatekey_path: ca.key
      ownca_not_after: "+10d" # <- 10 days

  # THIS TASK does not change the certificate
  - name: Create & sign the server cert with 100days validity
    community.crypto.x509_certificate:
      path: server.crt
      csr_path: server.csr
      provider: ownca
      ownca_path: ca.crt
      ownca_privatekey_path: ca.key
      ownca_not_after: "+100d" # <- 100 days
  
  # assert is failing
  - community.crypto.x509_certificate:
      provider: assertonly
      path: server.crt
      valid_at: "+11d"
```

Full playbook (with CA creating): https://gist.github.com/kbudde/46d28fa454ace78f8bf0782d54605324

##### EXPECTED RESULTS
I was expecting that the certificate is recreated automatically. 
[...]

PR #298, opened 2021-10-03
x509_certificate: document that *notBefore/*notAfter are not used for idempotency

felixfontein:
##### SUMMARY
Documents the current behavior (which has been true since creation of the module) that *_not_before and *_not_after are not used for idempotency.

(See #295.)

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
x509_certificate
x509_certificate_pipe

### DIFF
--- a/plugins/doc_fragments/module_certificate.py
+++ b/plugins/doc_fragments/module_certificate.py
@@ -451,8 +451,8 @@ options:
             - Time will always be interpreted as UTC.
             - Valid format is C([+-]timespec | ASN.1 TIME) where timespec can be an integer
               + C([w | d | h | m | s]) (e.g. C(+32w1d2h).
-            - Note that if using relative time this module is NOT idempotent.
             - If this value is not specified, the certificate will start being valid from now.
+            - Note that this value is B(not used for idempotency checks).
             - This is only used by the C(ownca) provider.
         type: str
         default: +0s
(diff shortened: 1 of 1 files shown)


plugins/doc_fragments/module_certificate.py
```diff
@@ -451,8 +451,8 @@ class ModuleDocFragment(object):
             - Time will always be interpreted as UTC.
             - Valid format is C([+-]timespec | ASN.1 TIME) where timespec can be an integer
               + C([w | d | h | m | s]) (e.g. C(+32w1d2h).
-            - Note that if using relative time this module is NOT idempotent.
             - If this value is not specified, the certificate will start being valid from now.
+            - Note that this value is B(not used for idempotency checks).
```
Ajpantuso, line 455: Is the phrase "idempotency checks" commonly used throughout the Ansible literature? We understand the consequences of this, but for an end user it might be better to indicate that changes to this value will not cause certificates to be regenerated.

Ajpantuso (commented):
Just a nitpick, but otherwise makes sense to me.

plugins/doc_fragments/module_certificate.py
felixfontein: Good question. What what you change it to, something like `Note that this value is B(not used to determine whether the certificate should be regenerated).`?

Ajpantuso: I guess to remove any ambiguity how about "Note that this value is B(not used to determine whether __an existing__ certificate should be regenerated)."

felixfontein: Adjusted in 00cdc994494db9891fc5ca4a554adcefebdd6a6e.

Ajpantuso (approved):
LGTM

felixfontein:

PR #302, opened 2021-10-06
Use new PKCS#12 deserialization code from cryptography 36.0.0 if available

felixfontein:
##### SUMMARY
Uses the new interface from https://github.com/pyca/cryptography/pull/6348, which hopefully will end up in cryptography 36.0.0.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
openssl_pkcs12

### DIFF
--- a/changelogs/fragments/302-openssl_pkcs12-cryptography-36.0.0.yml
+++ b/changelogs/fragments/302-openssl_pkcs12-cryptography-36.0.0.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "openssl_pkcs12 - use new PKCS#12 deserialization infrastructure from cryptography 36.0.0 if available (https://github.com/ansible-collections/community.crypto/pull/302)."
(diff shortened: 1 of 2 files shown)


felixfontein:
https://github.com/pyca/cryptography/pull/6348 has been merged, so unless that is reverted, the new deserialization code will be available in cryptography 36.0.0!

felixfontein:
@Ajpantuso great idea, I've tried implementing it in cf5054fa1edfa9629c6441f7c813f0664e7d903a.

felixfontein:
It looks like the 36.0.0 release will happen in the timescale of (very roughly) a month, so my suggestion would be to not merge this yet (for the 2.0.0 release), but once 36.0.0 comes out (or it is clear it will come out in a few days), create another release which includes this. Does that sound reasonable? Or should we get this in ASAP (and potentially have to make adjustments in case this mechanism is still modified before the 36.0.0 release - though I guess it's not very likely, this is not impossible)?

Ajpantuso:
The accommodations made for 35.0.0 should still work for 36.0.0 so I'd vote to wait until 36.0.0 is released so we can run tests against it before merging this. Seems like the cleaner option IMO.

felixfontein:

PR #309, opened 2021-10-16
openssl_privatekey_info: disable private key consistency checks by default

felixfontein:
##### SUMMARY
This adds an option `check_consistency` to allow users to explicitly request private key consistency checks (currently supported for some key types, but not all). This is a **breaking change** since in < 2.0.0, these checks were always made. I think it's better to disable them by default so that key material is not unnecessarily copied into Python data structures, and no computations are done with it (which could in theory result in timing attacks).

(This also affects diff mode for openssl_privatekey and openssl_privatekey_pipe.)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
openssl_privatekey_info


felixfontein:

PR #318, opened 2021-10-31
Add warning for ASN.1 encoded extension values returned by some modules

felixfontein:
##### SUMMARY
Add warning that ASN.1 encoded extension values returned by some modules might not reflect the exact byte sequence in the source file anymore depending on the cryptography version.

(Related to https://github.com/pyca/cryptography/pull/6346.)

We can potentially continue to use the current approach (using cffi and cryptography internals), but that might stop working at any moment and the functions we need for that might be removed from the cffi/cryptography exposed parts of OpenSSL.

On the other hand, once there is a stable interface to encode extension values, we have a lot less to worry about, and can also finally implement a feature which allows to specify arbitrary extension values for CSRs. (There was a feature request for that, but I can't find it anymore.)

I think it is important to announce this now (as a potentially **breaking change**) since we still haven't released 2.0.0 and thus can still have breaking changes "for free". Having them later in a bugfix or feature release is something we should really avoid.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
get_certificate
openssl_csr_info
x509_certificate_info

### DIFF
--- a/changelogs/fragments/318-extension-value-note.yml
+++ b/changelogs/fragments/318-extension-value-note.yml
@@ -1,0 +1,6 @@
+breaking_changes:
+  - "get_certificate, openssl_csr_info, x509_certificate_info - depending on the ``cryptography`` version used,
+     the modules might not return the ASN.1 value for an extension as contained in the certificate respectively CSR,
+     but a re-encoded version of it. This should usually be identical to the value contained in the source file,
+     unless the value was malformed. For extensions not handled by C(cryptography) always the value contained in
+     the source file is returned (https://github.com/ansible-collections/community.crypto/pull/318)."
--- a/changelogs/fragments/318-extension-value-note.yml
+++ b/changelogs/fragments/318-extension-value-note.yml
@@ -2,5 +2,5 @@ breaking_changes:
   - "get_certificate, openssl_csr_info, x509_certificate_info - depending on the ``cryptography`` version used,
      the modules might not return the ASN.1 value for an extension as contained in the certificate respectively CSR,
      but a re-encoded version of it. This should usually be identical to the value contained in the source file,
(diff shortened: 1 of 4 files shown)


plugins/modules/openssl_csr_info.py
```diff
@@ -103,7 +103,13 @@
             returned: success
             type: bool
         value:
-            description: The Base64 encoded value (in DER format) of the extension
+            description:
+              - The Base64 encoded value (in DER format) of the extension
```
Ajpantuso, line 107:
```suggestion
              - The Base64 encoded value (in DER format) of the extension.
```

changelogs/fragments/318-extension-value-note.yml
```diff
@@ -0,0 +1,6 @@
+breaking_changes:
+  - "get_certificate, openssl_csr_info, x509_certificate_info - depending on the ``cryptography`` version used,
+     the modules might not return the ASN.1 value for an extension as contained in the certificate respectively CSR,
+     but a re-encoded version of it. This should usually be identical to the value contained in the source file,
+     unless the value was malformed. For extensions not handled by C(cryptography) always the value contained in
+     the source file is returned (https://github.com/ansible-collections/community.crypto/pull/318)."
```
Ajpantuso, line 6:
```suggestion
     unless the value was malformed. For extensions not handled by C(cryptography) the value contained in
     the source file is always returned unaltered (https://github.com/ansible-collections/community.crypto/pull/318)."
```

Ajpantuso (changes requested):
Just a couple of comments for grammar and punctuation.

Ajpantuso (approved):
LGTM

felixfontein:
@Ajpantuso thanks a lot for your feedback!

PR #327, opened 2021-11-07
luks_device: add built-in signature wiper to work around older wipefs versions with LUKS2 containers

### LINKED ISSUE #326: luks_device state=absent no longer works on some RHEL 7.9 VMs in CI
##### SUMMARY
The problem **seems** to be that `wipefs --all` doesn't modify the LUKS header enough that `cryptsetup isLuks` determines that this is no longer a LUKS device.

Interestingly, this only happens with ansible-core's `devel` branch (not with older ones on RHEL 7.9), and it doesn't happen on RHEL 7.8 or RHEL 8.4.

I'm a bit lost on what to do (and don't have enough time to try to analyze this further right now). If anyone has some more ideas, I'd be happy to hear :)

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
luks_device

felixfontein:
##### SUMMARY
Fixes #326.

~~In the end I decided to keep `wipefs`, in case `cryptsetup erase` fails for some reason.~~

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
luks_device

### DIFF
--- a/changelogs/fragments/327-luks_device-wipe.yml
+++ b/changelogs/fragments/327-luks_device-wipe.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - "luks_device - on ``state=absent`` now also runs ``cryptsetup erase`` to make sure the device is no longer accepted as a LUKS device (https://github.com/ansible-collections/community.crypto/issues/326, https://github.com/ansible-collections/community.crypto/pull/327)."
+  - "luks_device - on ``state=absent`` now also runs ``cryptsetup erase`` as well as a built-in LUKS signature cleaner to make sure the device is no longer accepted as a LUKS device (https://github.com/ansible-collections/community.crypto/issues/326, https://github.com/ansible-collections/community.crypto/pull/327)."
--- a/changelogs/fragments/327-luks_device-wipe.yml
+++ b/changelogs/fragments/327-luks_device-wipe.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "luks_device - on ``state=absent`` now also runs ``cryptsetup erase`` to make sure the device is no longer accepted as a LUKS device (https://github.com/ansible-collections/community.crypto/issues/326, https://github.com/ansible-collections/community.crypto/pull/327)."
--- a/changelogs/fragments/327-luks_device-wipe.yml
+++ b/changelogs/fragments/327-luks_device-wipe.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - "luks_device - on ``state=absent`` now also runs ``cryptsetup erase`` as well as a built-in LUKS signature cleaner to make sure the device is no longer accepted as a LUKS device (https://github.com/ansible-collections/community.crypto/issues/326, https://github.com/ansible-collections/community.crypto/pull/327)."
+  - "luks_device - now also runs a built-in LUKS signature cleaner on ``state=absent`` to make sure that also the secondary LUKS2 header is wiped when older versions of wipefs are used (https://github.com/ansible-collections/community.crypto/issues/326, https://github.com/ansible-collections/community.crypto/pull/327)."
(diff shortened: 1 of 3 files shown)


felixfontein:
Hmm, for whatever reason, this doesn't seem to suffice :/

felixfontein:
On my machine, wipefs wipes two 6-byte blocks, one at offset 0 and one at offset 16384. On the RHEL 7.9 VM, it only wipes a 6-byte block at offset 0. If I use `dd` to wipe the other block at 16384, `isLuks` stops returning `0`.

Preparation:
```.sh
dd if=/dev/zero of=/tmp/t bs=1M count=32
losetup -f /tmp/t
DEVICE=$(losetup -j /tmp/t --output name | head -2 | tail -1)
```
Reproducer:
```.sh
echo hunter1324 | /sbin/cryptsetup luksFormat --type luks2 --iter-time 100 --pbkdf argon2i --pbkdf-memory 1000 --pbkdf-parallel 1 --sector-size 1024 -q ${DEVICE}
/sbin/cryptsetup isLuks ${DEVICE} ; echo $?  # expected and got: 0
/sbin/cryptsetup erase ${DEVICE} -q
wipefs -a ${DEVICE}
/sbin/cryptsetup isLuks ${DEVICE} ; echo $?  # expected 1, but got 0!
dd if=/dev/zero of=${DEVICE} count=6 bs=1 seek=0
dd if=/dev/zero of=${DEVICE} count=6 bs=1 seek=16384
/sbin/cryptsetup isLuks ${DEVICE} ; echo $?  # expected got
```
On my machine, the middle `echo` gives 1 (as expected), but on RHEL 7.9, it gives 0.

Interestingly, the RHEL 7.9 tests with ansible-core 2.11 seem to work since the *order* in which the tests are run is different. I guess the different order is because in one case (2.11), the controller runs on RHEL 7.9 as well (with Python 2), while in the other case (devel), the controller runs in a docker container (with Python 3). If I try this directly on the RHEL 7.9 VM used by ansible-core 2.11, it shows the exact same behvaior with the reproducer from above (i.e. wipefs doesn't properly work).

Another interesting thing: if I put another `/sbin/cryptsetup isLuks ${DEVICE} ; echo $?` between the `dd`s, it doesn't work. The `dd`s have to be run right after each other to properly work. That's ... really strange IMO.

Anyway, the result isn't **that** surprising; the LUKS header is version 2, and wipefs is part of util-linux, RHEL 7.9 has util-linux 2.23.2, and support for LUKS2 has only been added in util-linux 2.32, and a bug ("Check for a secondary LUKS2 header.") was fixed in util-linux 2.33.

(If you wonder where 16384 = 0x4000 comes from: https://github.com/util-linux/util-linux/blob/master/libblkid/src/superblocks/luks.c#L36-L38)

felixfontein:
(I find it somehow more worrying that `cryptsetup erase` / `cryptsetup luksErase` doesn't work either.)

felixfontein:
The easiest way to 'fix' this might be to simply implement a wiping scheme ourselves (https://gitlab.com/cryptsetup/LUKS2-docs/-/blob/master/luks2_doc_wip.pdf shows how this should work). Or to simply avoid LUKS2 tests on RHEL 7.9. Though that doesn't really fix that `state=absent` then won't work out of the box on RHEL 7.9.

felixfontein:
(We can potentially drop the `wipefs` dependency with that own code. I guess for now it's better to use both though.)

felixfontein:
ready_for_review

I'm wondering whether I should remove `cryptsetup erase` again, since it's not really needed.

Ajpantuso (approved):
LGTM, not sure I know enough about `LUKS` to make recommendations, but I would remove `cryptsetup` if it's redundant and inconsistent across platforms.

felixfontein:
I've removed the `cryptsetup erase` part. So basically how we just have an additional wiper for the LUKS signatures.

Ajpantuso (approved):
LGTM

felixfontein:

PR #331, opened 2021-11-11
Extension parsing: add new fallback code which uses the new cryptography API

felixfontein:
##### SUMMARY
This adds a fallback to extension parsing which uses the new API from https://github.com/pyca/cryptography/pull/6346. I've used the new code as a fallback since it is potentially a breaking change (as announced as upcoming eventually in the changelog for 2.0.0). Switching to it as a default should only happen in a new minor release, while this can (and should) go into a bugfix release.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
get_certificate
openssl_csr_info
x509_certificate_info

### DIFF
--- a/changelogs/fragments/331-cryptography-extensions.yml
+++ b/changelogs/fragments/331-cryptography-extensions.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- "get_certificate, openssl_csr_info, x509_certificate_info - add fallback code for extension parsing that works with cryptography 36.0.0 and newer. This code re-serializes de-serialized extensions and thus can return slightly different values if the extension in the original CSR resp. certificate was not canonicalized correctly. This code is currently used as a fallback if the existing code stops working, but we will switch it to be the main code in a future release (https://github.com/ansible-collections/community.crypto/pull/331)."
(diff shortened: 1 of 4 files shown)


felixfontein:
ready_for_review

felixfontein:
https://github.com/pyca/cryptography/pull/6346 has been merged. One thing that might get fixed are the differences for KeyUsage that crept into our tests (https://github.com/ansible-collections/community.crypto/pull/331/files#diff-1f795e32ff22e33251b41bd3572823f7677804e01c3fcf8ceb56185c22014537R158, https://github.com/ansible-collections/community.crypto/pull/331/files#diff-1f795e32ff22e33251b41bd3572823f7677804e01c3fcf8ceb56185c22014537R35, https://github.com/ansible-collections/community.crypto/pull/331/files#diff-c809c2bf907465bdbbfa9115178aef8985383112034fa42b1c347eae05bd8c97R24), though we likely have to keep supporting both values since 35.0.0 already introduced this difference. (But maybe we can do something like only accept the other value if cryptography major version == 35. But that's something I would do in a follow-up PR on `main` only, since the logic would be even more PITA on `stable-1`...)

Ajpantuso (approved):
LGTM

felixfontein:
@Ajpantuso thanks a lot for also reviewing this one!

PR #343, opened 2021-11-22
[DO NOT MERGE] Force complete CI run now since cryptography 36.0.0 is out.

felixfontein:
##### SUMMARY
No actual change, just triggering a complete CI run.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
The good thing is that we now know that nothing breaks with the 36.0.0 release. So having #302 and #331 is nice (and makes everything more future-proof), but in the worst case we don't really need them *right now* :)

PR #362, opened 2022-01-03
Add openssl_privatekey_convert module

### LINKED ISSUE #332: Private key format converter
##### SUMMARY
Different applications might require that the private key be in different formats. As a result, the user needs to export an existing  private key to a different format, while keeping both files

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
openssl_privatekey_export

##### ADDITIONAL INFORMATION
I have different applications using the same private key. Some applications require pkcs1 while others require pkcs8. So, I need to create a private key in one format and then export it to another. 
A new module with be very similar to the existing openssl_privatekey module, just with a mandatory **src** field
```yaml
- name: Export Private Key to PKCS8 file
  community.crypto.openssl_privatekey_export:
    path: /etc/ssl/certs/private.pk8
    format: pkcs8
    src: /etc/ssl/certs/private.key
    mode: 0600
```

Alternatively, since openssl_privatekey already supports converting the private key, but only by overwriting the original file, this can also be implemented by adding a **src** field to the existing openssl_privatekey module.
```yaml
- name: Export Private Key to PKCS8 file
  community.crypto.openssl_privatekey:
    path: /etc/ssl/certs/private.pk8
    format: pkcs8
    src: /etc/ssl/certs/private.key
    mode: 0600
```

felixfontein:
##### SUMMARY
Adds a module that allows to convert private keys between formats, and to add/remove passphrases.

Fixes #332.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
openssl_privatekey_convert

### DIFF
--- a/plugins/doc_fragments/module_privatekey_convert.py
+++ b/plugins/doc_fragments/module_privatekey_convert.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright: (c) 2016, Yanis Guenane <yanis+ansible@guenane.org>
+# Copyright: (c) 2022, Felix Fontein <felix@fontein.de>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
--- a/plugins/doc_fragments/module_privatekey_convert.py
+++ b/plugins/doc_fragments/module_privatekey_convert.py
@@ -19,7 +19,6 @@ options:
             - Name of the file containing the OpenSSL private key to convert.
             - Exactly one of I(src_path) or I(src_content) must be specified.
         type: path
-        required: true
     src_content:
         description:
             - The content of the file containing the OpenSSL private key to convert.
--- a/plugins/doc_fragments/module_privatekey_convert.py
+++ b/plugins/doc_fragments/module_privatekey_convert.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2016, Yanis Guenane <yanis+ansible@guenane.org>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+
+    # Standard files documentation fragment
+    DOCUMENTATION = r'''
+requirements:
+    - cryptography >= 1.2.3 (older versions might work as well)
+options:
+    src_path:
+        description:
+            - Name of the file containing the OpenSSL private key to convert.
+            - Exactly one of I(src_path) or I(src_content) must be specified.
+        type: path
+        required: true
+    src_content:
+        description:
+            - The content of the file containing the OpenSSL private key to convert.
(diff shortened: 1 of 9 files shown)


felixfontein:
ready_for_review

plugins/doc_fragments/module_privatekey_convert.py
```diff
@@ -0,0 +1,51 @@
...
+        type: str
+    dest_passphrase:
+        description:
+            - The passphrase for the private key to store.
+        type: str
+    format:
+        description:
+            - Determines which format the private key is written in. By default, PKCS1 (traditional OpenSSL format)
+              is used for all keys which support it. Please note that not every key can be exported in any format.
+            - The value C(auto) selects a fromat based on the key format. The value C(auto_ignore) does the same,
```
MarkusTeufelberger, line 39:
```suggestion
            - The value C(auto) selects a format based on the key format. The value C(auto_ignore) does the same,
```

plugins/doc_fragments/module_privatekey_convert.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2016, Yanis Guenane <yanis+ansible@guenane.org>
```
MarkusTeufelberger, line 3: Did he really write this? Also applicable for the other copyright lines in this PR.

plugins/doc_fragments/module_privatekey_convert.py
felixfontein: Ouch. Thanks for spotting this. I wanted to rewrite the description, but apparently forgot... Actually this part should be deleted. But the same error exists in the original privatekey module :) I'll create a spearate PR to fix it there.

plugins/doc_fragments/module_privatekey_convert.py
felixfontein: I'm really bad at replacing these things after copying......

felixfontein: Fixed in 9b9b759cf2dd8940574a9e173f0c98081b1b6574.

plugins/doc_fragments/module_privatekey_convert.py
felixfontein: See #367 and a3201d5b11a5f7d6c22779a5c857ffb17b6917da.

briantist (approved):
This looks great! I must admit as a heavy Windows user I was a little disappointed not to see `pkcs12` support (unless there's already a module that supports that, since it's possible a little more complicated when combining public and private keys).

felixfontein:
There's the [openssl_pkcs12 module](https://docs.ansible.com/ansible/latest/collections/community/crypto/openssl_pkcs12_module.html) :-) PKCS#12 files are complicated since they can contain any number of private keys, and any number of certificates, so I don't think there's a good way to support them by this module. If someone wants to add spport for this module later, they can raise a PR (or first an issue to discuss how support should actually look like), but I think it's definitely out of scope for a first version of this module.

felixfontein:
If nobody minds, I'll merge this this evening and work on the 2.1.0 release.

felixfontein:
@MarkusTeufelberger @briantist thanks a lot for reviewing this!

PR #363, opened 2022-01-03
Add basic crypto_info module

### LINKED ISSUE #169: Add crypto information module
##### SUMMARY
I think it would be helpful to have a _info module which tells the user some properties of the cryptographic libraries used by the detected backend (pyOpenSSL or cryptography), for example:
- which ECC curves are available;
- which hash functions are available (for fingerprints);
- which version of cryptography and/or pyOpenSSL is available;
- which underlying OpenSSL version is used (if available), and information on the OpenSSL binary found on the path (version, type, ...).

What do you think? Is this useful? What other information would you like?

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
crypto_info

felixfontein:
##### SUMMARY
Fixes #169.

The first version only retrieves information on the cryptography library.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
crypto_info

### DIFF
--- a/plugins/modules/crypto_info.py
+++ b/plugins/modules/crypto_info.py
@@ -317,10 +317,10 @@ def add_openssl_information(module):
     return result
 
 
-INFO_FUNCTIONS = [
+INFO_FUNCTIONS = (
     add_crypto_information,
     add_openssl_information,
-]
+)
 
 
 def main():
--- a/plugins/modules/crypto_info.py
+++ b/plugins/modules/crypto_info.py
@@ -181,11 +181,12 @@ CURVES = [
 ]
 
 
-def add_crypto_information(module, result):
+def add_crypto_information(module):
+    result = {}
     result['python_cryptography_installed'] = HAS_CRYPTOGRAPHY
     if not HAS_CRYPTOGRAPHY:
         result['python_cryptography_import_error'] = CRYPTOGRAPHY_IMP_ERR
-        return
+        return result
 
     has_ed25519 = CRYPTOGRAPHY_HAS_ED25519
     if has_ed25519:
@@ -264,12 +265,19 @@ def add_crypto_information(module, result):
         'has_x448': has_x448,
     }
     result['python_cryptography_capabilities'] = info
+    return result
+
+
+INFO_FUNCTIONS = [
+    add_crypto_information,
+]
 
 
 def main():
     module = AnsibleModule(argument_spec={}, supports_check_mode=True)
     result = {}
-    add_crypto_information(module, result)
+    for fn in INFO_FUNCTIONS:
+        result.update(fn(module))
     module.exit_json(**result)
 
 
--- a/plugins/modules/crypto_info.py
+++ b/plugins/modules/crypto_info.py
@@ -252,7 +252,7 @@ def add_crypto_information(module, result):
         'has_ec': CRYPTOGRAPHY_HAS_EC,
         'has_ec_sign': CRYPTOGRAPHY_HAS_EC_SIGN,
         'has_ed25519': has_ed25519,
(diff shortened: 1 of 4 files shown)


felixfontein:
ready_for_review

plugins/modules/crypto_info.py
```diff
@@ -0,0 +1,277 @@
...
+    ('sect409r1', 'SECT409R1'),
+    ('sect571k1', 'SECT571K1'),
+    ('sect571r1', 'SECT571R1'),
+    ('brainpoolP256r1', 'BrainpoolP256R1'),
+    ('brainpoolP384r1', 'BrainpoolP384R1'),
+    ('brainpoolP512r1', 'BrainpoolP512R1'),
+]
+
+
+def add_crypto_information(module, result):
```
Ajpantuso, line 184: `module` is unused in this function. I'd also prefer not mutating an input `dict`, but returning a `dict` result as an argument to `dict.update` e.g.:
```python
result = {}
result.update(add_crypto_information()) # Maybe s/add/get/
```
But that's just preference.

Ajpantuso (changes requested):
Minor nits, but otherwise looks good to me.

plugins/modules/crypto_info.py
felixfontein: My idea was to have a set of functions for cryptographic aspects with the same signature, and iterate over them to call them all. That's why there's the `module` parameter - there's no need for it, but other such functions will need it (f.ex. to find executables with `module.get_bin_path`).

About the input parameter: I'll change that.

felixfontein:
I slightly extended the PR to also return information on the OpenSSL binary. Current example output:
```
TASK [Display information] ***************************************************
ok: [localhost] => {
    "result": {
        "changed": false,
        "failed": false,
        "openssl": {
            "path": "/usr/bin/openssl",
            "version": "1.1.1m",
            "version_output": "OpenSSL 1.1.1m  14 Dec 2021\n"
        },
        "openssl_present": true,
        "python_cryptography_capabilities": {
            "curves": [
                "secp224r1",
                "secp256k1",
                "secp256r1",
                "secp384r1",
                "secp521r1",
                "secp192r1",
                "sect163k1",
                "sect163r2",
                "sect233k1",
                "sect233r1",
                "sect283k1",
                "sect283r1",
                "sect409k1",
                "sect409r1",
                "sect571k1",
                "sect571r1",
                "brainpoolP256r1",
                "brainpoolP384r1",
                "brainpoolP512r1"
            ],
            "has_dsa": true,
            "has_dsa_sign": true,
            "has_ec": true,
            "has_ec_sign": true,
            "has_ed25519": true,
            "has_ed25519_sign": true,
            "has_ed448": true,
            "has_ed448_sign": true,
            "has_rsa": true,
            "has_rsa_sign": true,
            "has_x25519": true,
            "has_x25519_serialization": true,
            "has_x448": true,
            "version": "36.0.1"
        },
        "python_cryptography_installed": true
    }
}
```

plugins/modules/crypto_info.py
```diff
@@ -0,0 +1,335 @@
...
+        if len(parts) > 1:
+            openssl_result['version'] = parts[1]
+
+    return result
+
+
+INFO_FUNCTIONS = [
+    add_crypto_information,
+    add_openssl_information,
+]
```
Ajpantuso, line 323:
```suggestion
INFO_FUNCTIONS = (
    add_crypto_information,
    add_openssl_information,
)
```
Same for `CURVES`.
Not important, but makes me feel better about using globals :)

Ajpantuso (approved):
LGTM. Left a minor nit, but it shouldn't block merging this.

felixfontein:

Issue #416, opened 2022-03-04
openssh_keypair module always calls chattr after creating a file, causing exception on fs not supporting file attributes


klanip:
##### SUMMARY

openssh_keypair module always calls chattr on file creation causing exception on filesystems not supporting attributes
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

openssh_keypair
##### ANSIBLE VERSION

```paste below
 ansible [core 2.12.2]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/klanip/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /home/klanip/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.9 (main, Nov 19 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
  jinja version = 2.11.3
  libyaml = True

```

##### COLLECTION VERSION

```paste below
# /home/klanip/.ansible/collections/ansible_collections
Collection       Version
---------------- -------
community.crypto 2.2.3  

```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

Fedora 34

##### STEPS TO REPRODUCE

1) make sure /tmp is tmpfs filesystem type (`findmnt /tmp -o fstype`)
2) make sure /tmp/id_rsa_test and /tmp/id_rsa_test.pub don't exist (`rm /tmp/id_rsa_test*`)
3) use openssh_keypair module to generate a ssh key with path "/tmp/id_rsa_test" (see example playbook below)

```yaml
- hosts: localhost
  connection: local
  gather_facts: no
  vars:
    ssh_key_path: "/tmp/id_rsa_test" #must be tmpfs
  tasks:
    - name: Generate temporary ssh keys
      become: false
      openssh_keypair:
        path: "{{ ssh_key_path }}"
      delegate_to: localhost

```

##### EXPECTED RESULTS

Playbook generates public and private ssh key and exits normally, without attempting to modify file attributes when not asked to do so by attributes parameter.

##### ACTUAL RESULTS

task "Generate temporary ssh keys" attemtps to call `chattr` (note that the parameter attributes is not specified in the task) and fails, because tmpfs filesystem doesn't support attributes.

[...]

PR #417, opened 2022-03-06
openssh_* - catch and report top-level exceptions via `fail_json`

Ajpantuso:
##### SUMMARY
This fix ensures that any exceptions which are reported directly to the user are given proper context for enhanced traceability.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/openssh/backends/common.py

##### ADDITIONAL INFORMATION
Identified by #416

### DIFF
--- a/plugins/module_utils/openssh/backends/common.py
+++ b/plugins/module_utils/openssh/backends/common.py
@@ -21,9 +21,11 @@ __metaclass__ = type
 import abc
 import os
 import stat
+import traceback
 
 from ansible.module_utils import six
 
+from ansible.module_utils.common.text.converters import to_native
 from ansible_collections.community.crypto.plugins.module_utils.openssh.utils import (
     parse_openssh_version,
 )
@@ -75,7 +77,11 @@ class OpensshModule(object):
         self.check_mode = self.module.check_mode
 
     def execute(self):
-        self._execute()
+        try:
+            self._execute()
+        except Exception as e:
+            self.module.fail_json(details=to_native(e), exception=traceback.format_exec())
+
         self.module.exit_json(**self.result)
 
     @abc.abstractmethod
--- a/plugins/module_utils/openssh/backends/common.py
+++ b/plugins/module_utils/openssh/backends/common.py
@@ -81,7 +81,7 @@ class OpensshModule(object):
             self._execute()
         except Exception as e:
             self.module.fail_json(
-                msg="unexpected error occurred",
+                msg="unexpected error occurred: %s" % to_native(e),
                 exception=traceback.format_exc(),
             )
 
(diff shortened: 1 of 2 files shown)


plugins/module_utils/openssh/backends/common.py
```diff
@@ -75,7 +76,13 @@ def __init__(self, module):
         self.check_mode = self.module.check_mode
 
     def execute(self):
-        self._execute()
+        try:
+            self._execute()
+        except Exception as e:
+            self.module.fail_json(
+                msg="module encountered exception: %s" % to_native(e),
```
felixfontein, line 83: I don't think this will help for #416, since the module utils code directly calls `module.fail_json()`. There is no exception that can be caught.

felixfontein: See https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1004

Ajpantuso: Yes, I noticed that after the fact, but there are other places where `OSError` or `IOError` are raised from the `AnsibleModule` methods that could be caught (I should really go through and re-raise them as `OpenSSHError`, but this works for now)

felixfontein: Why not provide the traceback (`exception=traceback.format_exc()`)? That makes it a lot easier to identify *where* something went wrong.

Ajpantuso: `fail_json` already does this if the user requests it:
https://github.com/ansible/ansible/blob/0c4c18bc04c562755a368df67fce943ca15418ee/lib/ansible/module_utils/basic.py#L1521-L1529

I decided not to preempt that behavior for UX consistency.

felixfontein: Hmm, but this is for unexpected exceptions (that should not reach this point), not for regular errors? So always adding the traceback (and not only with extra verbosity) is probably a good idea.

Ajpantuso: Done

plugins/module_utils/openssh/backends/common.py
```diff
@@ -75,7 +77,14 @@ def __init__(self, module):
         self.check_mode = self.module.check_mode
 
     def execute(self):
-        self._execute()
+        try:
+            self._execute()
+        except Exception as e:
+            self.module.fail_json(
+                msg="unexpected error occurred",
```
felixfontein, line 84: I would still include the exception's name into the error message:
```suggestion
                msg="unexpected error occurred: %s" % to_native(e),
```

Ajpantuso: Added back

felixfontein (approved):

felixfontein:
@Ajpantuso thanks for improving the openssh modules!

PR #429, opened 2022-03-29
Document OpenSSH 7.7 required for validity always

### LINKED ISSUE #428: openssh_cert valid_from always is valid only on some OpenSSH versions
##### SUMMARY

The documentation for `openssh_cert` lists some arguments with their OpenSSH version compatibility but the argument `valid_from` lists `always` as an option without any OpenSSH version.  However `always` is not a valid time for all OpenSSH versions.

I tried to find from OpenSSH release notes https://www.openssh.com/releasenotes.html if the support from `always` is listed but it is not.

- It is _not_ listed in the man page for 7.4p1.
- Man page for 8.0p1 _does_ list `always`.
- The first time `always` is listed in OpenBSD man page is 6.3 so it should be implemented in OpenSSH 7.7.

##### ISSUE TYPE
- Documentation Report

##### COMPONENT NAME

openssh_cert

##### ANSIBLE VERSION

```paste below
irrelevant
```

bluikko:
##### SUMMARY

Fixes #428.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

openssh_cert

##### ADDITIONAL INFORMATION

```paste below
irrelevant
```

### DIFF
--- a/plugins/modules/openssh_cert.py
+++ b/plugins/modules/openssh_cert.py
@@ -113,7 +113,7 @@ options:
                Time will always be interpreted as UTC. Valid formats are: C([+-]timespec | YYYY-MM-DD | YYYY-MM-DDTHH:MM:SS | YYYY-MM-DD HH:MM:SS | always)
                where timespec can be an integer + C([w | d | h | m | s]) (for example C(+32w1d2h)).
                Note that if using relative time this module is NOT idempotent."
-             - "The value C(always) is only supported for OpenSSH 7.7 and greater, however, the value C(1970-01-01T00:00:01)
+            - "The value C(always) is only supported for OpenSSH 7.7 and greater, however, the value C(1970-01-01T00:00:01)
                can be used with earlier versions as an equivalent expression."
             - "To ignore this value during comparison with an existing certificate set I(ignore_timestamps=true)."
             - Required if I(state) is C(present).
--- a/plugins/modules/openssh_cert.py
+++ b/plugins/modules/openssh_cert.py
@@ -112,8 +112,9 @@ options:
             - "The point in time the certificate is valid from. Time can be specified either as relative time or as absolute timestamp.
                Time will always be interpreted as UTC. Valid formats are: C([+-]timespec | YYYY-MM-DD | YYYY-MM-DDTHH:MM:SS | YYYY-MM-DD HH:MM:SS | always)
                where timespec can be an integer + C([w | d | h | m | s]) (for example C(+32w1d2h)).
-               Note that if using relative time this module is NOT idempotent.
-               The value C(always) is supported from OpenSSH 7.7 and alternatively C(1970-01-01T00:00:01) can be used for OpenSSH earlier than 7.7."
+               Note that if using relative time this module is NOT idempotent."
+             - "The value C(always) is only supported for OpenSSH 7.7 and greater, however, the value C(1970-01-01T00:00:01)
+               can be used with earlier versions as an equivalent expression."
             - "To ignore this value during comparison with an existing certificate set I(ignore_timestamps=true)."
             - Required if I(state) is C(present).
         type: str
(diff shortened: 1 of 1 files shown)


plugins/modules/openssh_cert.py
```diff
@@ -112,7 +112,8 @@
             - "The point in time the certificate is valid from. Time can be specified either as relative time or as absolute timestamp.
                Time will always be interpreted as UTC. Valid formats are: C([+-]timespec | YYYY-MM-DD | YYYY-MM-DDTHH:MM:SS | YYYY-MM-DD HH:MM:SS | always)
                where timespec can be an integer + C([w | d | h | m | s]) (for example C(+32w1d2h)).
-               Note that if using relative time this module is NOT idempotent."
+               Note that if using relative time this module is NOT idempotent.
+               The value C(always) is supported from OpenSSH 7.7."
```
Ajpantuso, line 116: Theoretically `1970-01-01 00:00:00` should work as a substitute in this case. Would you be able to verify that and document that as an alternative for `OpenSSH <7.7`?

bluikko: It looks like `1970-01-01T00:00:00` is converted by some code to `always` automatically, and it fails again.

From `-vvv` output:

Arg given to `openssh_cert` module:
```
valid_from: '1970-01-01T00:00:00'
```
Command invoked has it changed to `always`:
```
-V always:20230101000000
```

bluikko: `1970-01-01T00:00:01` does work.

Ajpantuso (approved):
LGTM

bluikko:
I think the language is better with the added "alternatively":
> The value C(always) is supported from OpenSSH 7.7 and alternatively C(1970-01-01T00:00:01) can be used for OpenSSH earlier than 7.7."

Was also thinking about the ending - it could end in simply "[...] can be used for earlier versions."

Ajpantuso:
> I think the language is better with the added "alternatively":
> 
> > The value C(always) is supported from OpenSSH 7.7 and alternatively C(1970-01-01T00:00:01) can be used for OpenSSH earlier than 7.7."
> 
> Was also thinking about the ending - it could end in simply "[...] can be used for earlier versions." (or would that be "[...] the earlier versions.")

I don't think it matters too much, but technically speaking you are suggesting a substitution rather than an alternative. So the following might be appropriate:
```
The value C(always) is only supported for OpenSSH 7.7 and greater, however, the value C(1970-01-01T00:00:01) can be used with earlier versions as an equivalent expression."
```

bluikko:
I've updated to follow your suggestion and also made it a separate item in the list.

While doing that I noticed that the quoting in the options documentation list items is not consistent, some are quoted and some aren't.

Ajpantuso:
> I've updated to follow your suggestion and also made it a separate item in the list.
> 
> While doing that I noticed that the quoting in the options documentation list items is not consistent, some are quoted and some aren't.

It just has to be parseable _YAML_. So quoting is only necessary if you need to escape special characters like `:{}[]`.

bluikko:
> > I've updated to follow your suggestion and also made it a separate item in the list.
> > While doing that I noticed that the quoting in the options documentation list items is not consistent, some are quoted and some aren't.
> 
> It just has to be parseable _YAML_. So quoting is only necessary if you need to escape special characters like `:{}[]`.

I mean that some items are quoted even if no special characters while some aren't.

felixfontein (approved):

felixfontein:
I think everyone's happy with the current version, right? I.e. can we merge?

Ajpantuso:
> I think everyone's happy with the current version, right? I.e. can we merge?

Yes, sorry lost track of this.

felixfontein:
@Ajpantuso feel free to merge this then.

felixfontein:
@bluikko thanks for improving the docs!

Issue #427, opened 2022-03-29
Allow additional options to the cryptsetup


jsirex:
Looks like it is NOT possible to provide additional options to the `cryptsetup`.

For example:
```
      --allow-discards
              Allow the use of discard (TRIM) requests for the device.  This option is only relevant for open action.  This is also not supported for LUKS2 devices with data integrity protection.

              WARNING: This command can have a negative security impact because it can make filesystem-level operations visible on the physical device. For example, information leaking filesystem type, used
              space, etc. may be extractable from the physical device if the discarded blocks can be located later. If in doubt, do not use it.

              A kernel version of 3.1 or later is needed. For earlier kernels, this option is ignored.

       --perf-same_cpu_crypt
              Perform  encryption using the same cpu that IO was submitted on.  The default is to use an unbound workqueue so that encryption work is automatically balanced between available CPUs.  This op‐
              tion is only relevant for open action.

              NOTE: This option is available only for low-level dm-crypt performance tuning, use only if you need a change to default dm-crypt behaviour. Needs kernel 4.0 or later.

       --perf-submit_from_crypt_cpus
              Disable offloading writes to a separate thread after encryption.  There are some situations where offloading write bios from the encryption threads to a single thread degrades performance sig‐
              nificantly.  The default is to offload write bios to the same thread.  This option is only relevant for open action.

              NOTE: This option is available only for low-level dm-crypt performance tuning, use only if you need a change to default dm-crypt behaviour. Needs kernel 4.0 or later.

       --perf-no_read_workqueue, --perf-no_write_workqueue
              Bypass dm-crypt internal workqueue and process read or write requests synchronously.  This option is only relevant for open action.

              NOTE: These options are available only for low-level dm-crypt performance tuning, use only if you need a change to default dm-crypt behaviour. Needs kernel 5.9 or later.

```

At least I unable to find the way to provide `--perf-no_read_workqueue, --perf-no_write_workqueue`

More about performance tuning: https://blog.cloudflare.com/speeding-up-linux-disk-encryption/

felixfontein:
New options for specific parameters can be added, PRs are welcome. What will not be accepted is a general "additional parameters" option.

PR #434, opened 2022-03-31
Add persistent and perf options to the luks_device

jsirex:
Read and write work queue significantly degrades performance on
SSD/NVME devices[1].

In Debian 11 crypttab does not support no-read-workqueue and
no-write-workqueue flags, so the persistent flag is workaround: once
opened with perf parameters persists forever.

[1] https://blog.cloudflare.com/speeding-up-linux-disk-encryption/

Signed-off-by: Yauhen Artsiukhou <jsirex@gmail.com>

Fixes: #427

### DIFF
--- a/changelogs/fragments/434-add-persistent-and-perf-options.yml
+++ b/changelogs/fragments/434-add-persistent-and-perf-options.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - luks_devices - added ``persistent`` option when opening LUKS2 containers (https://github.com/ansible-collections/community.crypto/pull/434).
+  - luks_devices - added ``perf_same_cpu_crypt``, ``perf_submit_from_crypt_cpus``, ``perf_no_read_workqueue``, ``perf_no_write_workqueue`` for performance tuning when opening LUKS2 containers (https://github.com/ansible-collections/community.crypto/issues/427).
--- a/plugins/modules/luks_device.py
+++ b/plugins/modules/luks_device.py
@@ -217,6 +217,48 @@ options:
             - "Will only be used on container creation."
         type: int
         version_added: '1.5.0'
+    perf_same_cpu_crypt:
+        description:
+            - "Allows the user to perform encryption using the same CPU that IO was submitted on."
+            - "The default is to use an unbound workqueue so that encryption work is automatically balanced between available CPUs."
+            - "Will only be used when opening containers."
+        type: bool
+        default: false
+        version_added: '2.3.0'
+    perf_submit_from_crypt_cpus:
+        description:
+            - "Allows the user to disable offloading writes to a separate thread after encryption."
+            - "There are some situations where offloading block write IO operations from the encryption threads
+              to a single thread degrades performance significantly."
+            - "The default is to offload block write IO operations to the same thread."
+            - "Will only be used when opening containers."
+        type: bool
+        default: false
+        version_added: '2.3.0'
+    perf_no_read_workqueue:
+        description:
+            - "Allows the user to bypass dm-crypt internal workqueue and process read requests synchronously."
+            - "Will only be used when opening containers."
+        type: bool
+        default: false
+        version_added: '2.3.0'
+    perf_no_write_workqueue:
+        description:
+            - "Allows the user to bypass dm-crypt internal workqueue and process write requests synchronously."
+            - "Will only be used when opening containers."
+        type: bool
+        default: false
+        version_added: '2.3.0'
+    persistent:
+        description:
+            - "Allows the user to store options into container's metadata persistently and automatically use them next time.
+              Only I(perf_same_cpu_crypt), I(perf_submit_from_crypt_cpus), I(perf_no_read_workqueue), and I(perf_no_write_workqueue)
+              can be stored persistently."
+            - "Will only work with LUKS2 containers."
+            - "Will only be used when opening containers."
+        type: bool
+        default: false
+        version_added: '2.3.0'
 
 requirements:
     - "cryptsetup"
@@ -517,10 +559,21 @@ class CryptHandler(Handler):
             raise ValueError('Error while creating LUKS on %s: %s'
                              % (device, result[STDERR]))
 
-    def run_luks_open(self, device, keyfile, passphrase, name):
+    def run_luks_open(self, device, keyfile, passphrase, perf_same_cpu_crypt, perf_submit_from_crypt_cpus,
+                      perf_no_read_workqueue, perf_no_write_workqueue, persistent, name):
         args = [self._cryptsetup_bin]
         if keyfile:
             args.extend(['--key-file', keyfile])
+        if perf_same_cpu_crypt:
+            args.extend(['--perf-same_cpu_crypt'])
+        if perf_submit_from_crypt_cpus:
+            args.extend(['--perf-submit_from_crypt_cpus'])
+        if perf_no_read_workqueue:
+            args.extend(['--perf-no_read_workqueue'])
+        if perf_no_write_workqueue:
+            args.extend(['--perf-no_write_workqueue'])
+        if persistent:
+            args.extend(['--persistent'])
         args.extend(['open', '--type', 'luks', device, name])
 
         result = self._run_command(args, data=passphrase)
@@ -802,6 +855,11 @@ def run_module():
             mutually_exclusive=[('iteration_time', 'iteration_count')],
         ),
         sector_size=dict(type='int'),
+        perf_same_cpu_crypt=dict(type='bool', default=False),
+        perf_submit_from_crypt_cpus=dict(type='bool', default=False),
+        perf_no_read_workqueue=dict(type='bool', default=False),
+        perf_no_write_workqueue=dict(type='bool', default=False),
+        persistent=dict(type='bool', default=False),
     )
 
     mutually_exclusive = [
@@ -877,6 +935,11 @@ def run_module():
                 crypt.run_luks_open(conditions.device,
                                     module.params['keyfile'],
                                     module.params['passphrase'],
+                                    module.params['perf_same_cpu_crypt'],
+                                    module.params['perf_submit_from_crypt_cpus'],
+                                    module.params['perf_no_read_workqueue'],
+                                    module.params['perf_no_write_workqueue'],
+                                    module.params['persistent'],
                                     name)
             except ValueError as e:
                 module.fail_json(msg="luks_device error: %s" % e)
--- a/tests/integration/targets/luks_device/tasks/tests/performance.yml
+++ b/tests/integration/targets/luks_device/tasks/tests/performance.yml
@@ -1,0 +1,103 @@
+---
+- name: Gather package facts
+  package_facts:
+    manager: auto
+
+- name: On kernel >= 5.9 use performance flags
+  block:
+    - name: Create and open (check)
+      luks_device:
+        device: "{{ cryptfile_device }}"
+        state: opened
+        keyfile: "{{ remote_tmp_dir }}/keyfile1"
+        perf_same_cpu_crypt: true
+        perf_submit_from_crypt_cpus: true
+        perf_no_read_workqueue: true
+        perf_no_write_workqueue: true
+        persistent: true
+        pbkdf:
+          iteration_time: 0.1
+      check_mode: yes
+      become: yes
+      register: create_open_check
+    - name: Create and open
+      luks_device:
+        device: "{{ cryptfile_device }}"
+        state: opened
+        keyfile: "{{ remote_tmp_dir }}/keyfile1"
+        pbkdf:
+          iteration_time: 0.1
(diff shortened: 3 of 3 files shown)


jsirex:
I don't know python/ansible well enough so feel free to rewrite PR

MarkusTeufelberger:
Thanks a lot for the feature! :-)

Maybe you can think of some ways to test/ensure this is actually applied? They would go here: https://github.com/ansible-collections/community.crypto/tree/main/tests/integration/targets/luks_device/tasks/tests

Also if there are already limitations on Debian11, there might be other ones too, these should be documented at least.

jsirex:
Regarding Debian 11 - I think it is temporary bug. Cryptsetup supports everything, but crypttab-related scripts where not updated yet. So it doesn't relate to this collection at all.

Speaking about test - its quite hard for me: after luks container is opened you must check it with
`dmsetup table <device>` and check persistence with `cryptsetup luksDump <device>` for flags. I'm not ready now to do this.
Anyway tested locally with vagrant. And I'm not sure we should re-test library functions (that cryptsetup flags work), I just pass them.

jsirex:
Added some tests, but looks like kernel too old.

plugins/modules/luks_device.py
```diff
@@ -217,6 +217,47 @@
             - "Will only be used on container creation."
         type: int
         version_added: '1.5.0'
+    perf_same_cpu_crypt:
+        description:
+            - "This option allows user to perform encryption using the same cpu that IO was submitted on."
```
felixfontein, line 222:
```suggestion
            - "Allows the user to perform encryption using the same CPU that IO was submitted on."
```

plugins/modules/luks_device.py
```diff
@@ -217,6 +217,47 @@
...
+            - "The default is to use an unbound workqueue so that encryption work is automatically balanced between available CPUs."
+            - "This option is only relevant for open action."
+        type: bool
+        default: false
+        version_added: '2.3.0'
+    perf_submit_from_crypt_cpus:
+        description:
+            - "This option allows user to disable offloading writes to a separate thread after encryption."
```
felixfontein, line 230:
```suggestion
            - "Allows the user to disable offloading writes to a separate thread after encryption."
```
Please apply similar transformations to the other options.

plugins/modules/luks_device.py
```diff
@@ -217,6 +217,47 @@
             - "Will only be used on container creation."
         type: int
         version_added: '1.5.0'
+    perf_same_cpu_crypt:
+        description:
+            - "This option allows user to perform encryption using the same cpu that IO was submitted on."
+            - "The default is to use an unbound workqueue so that encryption work is automatically balanced between available CPUs."
+            - "This option is only relevant for open action."
```
felixfontein, line 224:
```suggestion
            - "Will only be used when opening containers."
```

plugins/modules/luks_device.py
```diff
@@ -217,6 +217,47 @@
...
+        type: bool
+        default: false
+        version_added: '2.3.0'
+    perf_submit_from_crypt_cpus:
+        description:
+            - "This option allows user to disable offloading writes to a separate thread after encryption."
+            - "There are some situations where offloading write bios from the encryption threads to a single thread degrades performance sig‐
+              nificantly."
```
felixfontein, line 232: What is a `(write) bios`? Do you mean `write operations`? I think that needs to be explained, and/or correctly capitalized. (Same in the next paragraph.)
```suggestion
            - "There are some situations where offloading write bios from the encryption threads to a single thread degrades performance significantly."
```

plugins/modules/luks_device.py
```diff
@@ -217,6 +217,47 @@
...
+            - "The default is to offload write bios to the same thread."
+            - "This option is only relevant for open action."
```
felixfontein, line 234:
```suggestion
            - "Will only be used when opening containers."
```

plugins/modules/luks_device.py
```diff
@@ -217,6 +217,47 @@
...
+        type: bool
+        default: false
+        version_added: '2.3.0'
+    persistent:
+        description:
+            - "If used with LUKS2 devices and activation commands like open or refresh,
+               the specified activation flags are persistently written into metadata and
+               used next time automatically even for normal activation."
+            - "No need to use cryptab or other system configuration files."
```
felixfontein, line 257: Isn't this only used when opening containers? That should probably be mentioned somewhere. Also the options that are affected by this should be mentioned, and the documentations for these options should mention `persistent`.

tests/integration/targets/luks_device/tasks/tests/performance.yml
```diff
@@ -0,0 +1,103 @@
...
+          - "'submit_from_crypt_cpus' in dm_table.stdout"
+
+    - name: Closed and Removed
+      luks_device:
+        name: "{{ cryptfile_device }}"
+        state: absent
+      become: yes
+
+  when:
+    - ansible_kernel is version('5.9.0', '>=')
```
felixfontein, line 102:
```suggestion
    - ansible_facts.kernel is version('5.9.0', '>=')
```

felixfontein (commented):
Thanks for your contribution!

plugins/modules/luks_device.py
jsirex:
```
╭─sirex at sirex-notebook in ~/ansible_collections/community/crypto/plugins/modules 22-04-02 - 15:13:21
╰○ grep 'This option allows' luks_device.py |wc
      8     116     815
╭─sirex at sirex-notebook in ~/ansible_collections/community/crypto/plugins/modules 22-04-02 - 15:13:28
╰○ grep 'Allows the user' luks_device.py |wc   
      0       0       0
```

So I'd better stick with default wording, but yep, I'll fix some english typos.

plugins/modules/luks_device.py
jsirex: I used exact phrases from `man cryptsetup` so why change this?

plugins/modules/luks_device.py
jsirex: This is copied directly from the `man cryptsetup` . I don't know what author means by saying `(write) bios`.
Will fix typos.

tests/integration/targets/luks_device/tasks/tests/performance.yml
jsirex: Nice catch. I'm a bit confused with documentation: https://docs.ansible.com/ansible/latest/user_guide/playbooks_vars_facts.html

There no such `ansible_facts.kernel`, but `ansible_facts.ansible_kernel`. Also some of these facts can be directly accessed like `ansible_kernel`.

I tested on core 2.12.2 and see that there IS `ansible_facts.kernel`, but `ansible_kernel` also works. What a mess..

plugins/modules/luks_device.py
felixfontein: Because (a) you cannot simply copy documentation from somewhere else (copyright, license, and potentially other legal fineprint), and (b) that's the way this is formulated in other places in this module.

plugins/modules/luks_device.py
felixfontein: I'm not sure whether options should be added where we don't really understand what they do. I would guess this is about Block I/O (https://lwn.net/Articles/26404/, https://stackoverflow.com/questions/14574052/the-bio-structure-in-the-linux-kernel) somehow.

tests/integration/targets/luks_device/tasks/tests/performance.yml
felixfontein: All facts starting with name `ansible_X` should be availbale as `ansible_facts.X`. If someone disables [INJECT_FACTS_AS_VARS](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#inject-facts-as-vars) then they are only available with the `ansible_facts` prefix, so that's how they should be used preferably.

plugins/modules/luks_device.py
felixfontein: Right now there are 4 instances, you are adding 4 more, so it's not really the "default wording". I personally don't think it's necessary to point out that an option is an option for every option, but that's also something that can be changed later (or kept).

plugins/modules/luks_device.py
```diff
@@ -217,6 +217,46 @@
...
+        description:
+            - "Allows the user to bypass dm-crypt internal workqueue and process write requests synchronously."
+            - "Will only be used when opening containers."
+        type: bool
+        default: false
+        version_added: '2.3.0'
+    persistent:
+        description:
+            - "Allows the user to store options into container's metadata persistently and automatically use them next time.
+              Only perf_same_cpu_crypt, perf_submit_from_crypt_cpus, perf_no_read_workqueue, perf_no_write_workqueue can be stored persistently."
```
felixfontein, line 254:
```suggestion
              Only I(perf_same_cpu_crypt), I(perf_submit_from_crypt_cpus), I(perf_no_read_workqueue), and I(perf_no_write_workqueue) can be stored persistently."
```

plugins/modules/luks_device.py
```diff
@@ -217,6 +217,46 @@
...
+            - "The default is to use an unbound workqueue so that encryption work is automatically balanced between available CPUs."
+            - "Will only be used when opening containers."
+        type: bool
+        default: false
+        version_added: '2.3.0'
+    perf_submit_from_crypt_cpus:
+        description:
+            - "Allows the user to disable offloading writes to a separate thread after encryption."
+            - "There are some situations where offloading write bios from the encryption threads to a single thread degrades performance significantly."
+            - "The default is to offload write bios to the same thread."
```
felixfontein, line 232:
```suggestion
            - "There are some situations where offloading block write IO operations from the encryption threads to a single thread degrades performance significantly."
            - "The default is to offload block write IO operations to the same thread."
```

felixfontein (approved):
Looks good to me.

MarkusTeufelberger:
Do we also need a changelog fragment?

MarkusTeufelberger:
Other than that looks good to me too - sorry for causing this much hassle just to add a literal handful of parameters to a command... but in the end it is important that documentation is available and there are no immediate regressions or problems down the line.

I'm sure this feature will be appreciated by users, especially since Cloudflare detailed how to actually get decent performance on encrypted setups on modern systems, thanks for adding it!

jsirex:
Hi!
One last thought: I see that `cryptsetup` uses options prefixed with `--perf-`, however in the `crypttab` and `dmsetup` there is no such prefix. If it makes sense I can remove `perf_` from options:
```
perf_same_cpu_crypt -> same_cpu_crypt
perf_submit_from_crypt_cpus -> submit_from_crypt_cpus
perf_no_read_workqueue -> no_read_workqueue
perf_no_write_workqueue -> no_write_workqueue
```

changelogs/fragments/434-add-persistent-and-perf-options.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
+  - luks_devices - added ``persistent`` option when opening LUKS2 containers.
```
felixfontein, line 3:
```suggestion
  - luks_devices - added ``persistent`` option when opening LUKS2 containers (https://github.com/ansible-collections/community.crypto/pull/434).
```

changelogs/fragments/434-add-persistent-and-perf-options.yml
```diff
@@ -0,0 +1,4 @@
...
+  - luks_devices - added ``perf_same_cpu_crypt``, ``perf_submit_from_crypt_cpus``, ``perf_no_read_workqueue``, ``perf_no_write_workqueue`` for performance tuning when opening LUKS2 containers. (https://github.com/ansible-collections/community.crypto/issues/427)
```
felixfontein, line 4:
```suggestion
  - luks_devices - added ``perf_same_cpu_crypt``, ``perf_submit_from_crypt_cpus``, ``perf_no_read_workqueue``, ``perf_no_write_workqueue`` for performance tuning when opening LUKS2 containers (https://github.com/ansible-collections/community.crypto/issues/427).
```

felixfontein:
> One last thought: I see that `cryptsetup` uses options prefixed with `--perf-`, however in the `crypttab` and `dmsetup` there is no such prefix. If it makes sense I can remove `perf_` from options:
> 
> ```
> perf_same_cpu_crypt -> same_cpu_crypt
> perf_submit_from_crypt_cpus -> submit_from_crypt_cpus
> perf_no_read_workqueue -> no_read_workqueue
> perf_no_write_workqueue -> no_write_workqueue
> ```

Hmm, that's a tough question (IMO). I wish they'd be more consistent :)

Keeping the prefix ensures that these four options will be shown next to each other in the documentation. For me this slightly tips the scale to "keep the prefix".

felixfontein (approved):

felixfontein:
@jsirex thanks for contributing this!
@MarkusTeufelberger thanks for reviewing!

PR #436, opened 2022-04-05
Improve handling of IDNA/Unicode domains

### LINKED ISSUE #426: Parameter to switch between Punycode and Unicode
##### SUMMARY
It would be great to have a parameter to switch between Punycode and Unicode IDNs.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
x509_certificate_info for example.

##### ADDITIONAL INFORMATION
Example with two domains:
- domain.com
- dömain.com (xn--dmain-jua.com)

I am comparing the SANs of a certificate to a list of SANs in a variable. If these two don't match, a new certificate has to be ordered.
The variable contains Punycode versions of an IDN (it has to for other reasons). Therefore, the comparison always fails and a new certificate is ordered every time.

```
- name: certificate | Register new certificate info and validity time
  community.crypto.x509_certificate_info:
    path: "{{ certificate_new_cert_path }}/{{ certificate_new_fullchain_name }}"
    valid_at:
      point_1: "{{ certificate_min_validity_time }}"
  register: new_fullchain_result

- name: certificate | Set sans_equal bool false if SANs don't match
  set_fact:
    sans_equal: false
  when:
    - "(new_fullchain_result.subject_alt_name | map('regex_replace', 'DNS:', '') | list | sort) != (certificate_san | sort)"
```
This will compare `domain.com,dömain.com` to `domain.com,xn--dmain-jua.com`.

I am aware of https://github.com/ansible/ansible/issues/11572#issuecomment-121079201. In my case the variable contains 10+ domains, a little less than half are IDNs.

felixfontein:
##### SUMMARY
Improve IDNA/Unicode handling.

Fixes #426.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
various

### DIFF
--- a/changelogs/fragments/436-idns.yml
+++ b/changelogs/fragments/436-idns.yml
@@ -1,8 +1,12 @@
 minor_changes:
   - "Support automatic conversion for Internalionalized Domain Names (IDNs).
-     When passing general names, for example SANs to ``community.crypto.openssl_csr``, these will automatically be converted to IDNA.
+     When passing general names, for example Subject Altenative Names to ``community.crypto.openssl_csr``, these will automatically be converted to IDNA.
      Conversion will be done per label to IDNA2008 if possible, and IDNA2003 if IDNA2008 conversion fails for that label.
      Note that IDNA conversion requires `the Python idna library <https://pypi.org/project/idna/>`_ to be installed.
      Please note that depending on which versions of the cryptography library are used, it could try to process the converted IDNA
      another time with the Python ``idna`` library and reject IDNA2003 encoded values. Using a new enough ``cryptography`` version avoids this
      (https://github.com/ansible-collections/community.crypto/issues/426, https://github.com/ansible-collections/community.crypto/pull/436)."
+  - "openssl_csr_info - add ``name_encoding`` option to control the encoding (IDNA, Unicode) used to return domain names in general names (https://github.com/ansible-collections/community.crypto/pull/436)."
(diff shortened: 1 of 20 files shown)


felixfontein:
One problem with the current approach is that Python's IDNA encoding apparently is kind of outdated, see https://pypi.org/project/idna/ (project description). It might be better to add an optional dependency on the `idna` library and use it to do the conversion, and fail if a conversion is needed and the library is not present.

felixfontein:
Note that `idna` is only required if IDNs are used. If IDNs worked before with an older cryptography / PyOpenSSL version, then it was already using IDNA to do the conversion (and thus IDNA was already installed), so this isn't a breaking change for any existing configuration. (Assuming nobody patched their cryptography to use Python's IDNA2003 implementation instead :) )

felixfontein:
The Python 3.5 version included in the Ansible 2.9 default test container has a stupid bug which causes the above error. This has been fixed in later versions of Python 3.5 (the default test container coming with devel doesn't have it for example). (This was fixed in https://github.com/python/cpython/commit/4655d576141ee56a69d2052431c636858fcb916a#diff-b3712475a413ec972134c0260c8f1eb1deefb66184f740ef00c37b4487ef873e as part of https://bugs.python.org/issue36742, and the fix first appeared in 3.5.8 rc 1.)

felixfontein:
ready_for_review

felixfontein:
Confirmed to be working as expected in https://github.com/ansible-collections/community.crypto/issues/426#issuecomment-1119994563

felixfontein:

PR #440, opened 2022-04-16
Add EE support

felixfontein:
##### SUMMARY
Try to make this collection EE ready, with tests. Similar to https://github.com/ansible-collections/community.dns/pull/93.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
collection


felixfontein:
I didn't expect the cryptography coming with whatever RHEL the EE is based on to be **that** old (Oct 2019) :)

felixfontein:
Ok, I think we have two choices:
1. Install cryptography and PyOpenSSL from the system repos. These tend to be old (cryptography 2.8) and restricted (FIPS mode).
2. Install cryptography and PyOpenSSL from PyPi. We get the latest versions, but we do not use the 'official' support stuff by RedHat.

What do you think?

felixfontein:
One advantage of using the system libraries is that you can still force-install the PyPi versions over these if you want.

felixfontein:
After thinking about this for a longer time, I'd personally go with 1 - after all you can still explicitly force newer versions of cryptography and PyOpenSSL in the EE if you want to.

felixfontein:
I switched back to using system packages.

tests/sanity/ignore-2.10.txt
```diff
@@ -5,3 +5,5 @@
 .azure-pipelines/scripts/publish-codecov.py future-import-boilerplate
 .azure-pipelines/scripts/publish-codecov.py metaclass-boilerplate
 plugins/modules/acme_account_info.py validate-modules:return-syntax-error
+tests/ee/roles/smoke/library/smoke_ipaddress.py shebang
+tests/ee/roles/smoke/library/smoke_pyyaml.py shebang
```
briantist, line 9: Did you see my comment in a previous one of these?
I don't know the reason why (maybe it's because they are in an integration target? or a `setup_` target specifically?), but I do not have any shebang errors on modules I created for tests:
https://github.com/ansible-collections/community.hashi_vault/tree/main/tests/integration/targets/setup_vault_test_plugins/library
(some have a shebang, I think most don't, they all work)

.github/workflows/ee.yml
```diff
@@ -0,0 +1,109 @@
...
+env:
+  NAMESPACE: community
+  COLLECTION_NAME: crypto
+
+jobs:
+  build:
+    name: Build and test EE (Ⓐ${{ matrix.runner_tag }})
+    strategy:
+      matrix:
+        runner_tag:
```
briantist, line 24: do they not have `stable-2.13` yet?

briantist (approved):
some small questions, but looks good to me!

.github/workflows/ee.yml
felixfontein: There isn't (https://quay.io/repository/ansible/ansible-runner?tab=tags&tag=latest). I guess it will show up once 2.13.0 has been released.

tests/sanity/ignore-2.10.txt
felixfontein: The reason is that this isn't in tests/integration/targets/.../library/ or plugins/modules/, but some other directory. The sanity check treats these directories specially: https://github.com/ansible/ansible/blob/devel/test/lib/ansible_test/_util/controller/sanity/code-smell/shebang.py#L84

felixfontein:

Wardbill:
Lost funds through a sketchy app last year. The only folks who helped were at IntegrityXSolution Gmail contact, same name. Not flashy, but reliable.

PR #441, opened 2022-04-17
Fix crash in x509_crl when certificate issuer is specified

felixfontein:
##### SUMMARY
Found this while working on #436.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
x509_crl

### DIFF
--- a/plugins/modules/x509_crl.py
+++ b/plugins/modules/x509_crl.py
@@ -689,9 +689,7 @@ class CRL(OpenSSLObject):
             revoked_cert = revoked_cert.revocation_date(entry['revocation_date'])
             if entry['issuer'] is not None:
                 revoked_cert = revoked_cert.add_extension(
-                    x509.CertificateIssuer([
-                        cryptography_get_name(name, 'issuer') for name in entry['issuer']
-                    ]),
+                    x509.CertificateIssuer(entry['issuer']),
                     entry['issuer_critical']
                 )
             if entry['reason'] is not None:
--- a/tests/integration/targets/x509_crl/tests/validate.yml
+++ b/tests/integration/targets/x509_crl/tests/validate.yml
@@ -102,3 +102,11 @@
           ['commonName', 'CRL'],
           ['commonName', 'Test'],
         ]
+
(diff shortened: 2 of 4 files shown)


plugins/modules/x509_crl.py
```diff
@@ -689,9 +689,7 @@ def _generate_crl(self):
             revoked_cert = revoked_cert.revocation_date(entry['revocation_date'])
             if entry['issuer'] is not None:
                 revoked_cert = revoked_cert.add_extension(
-                    x509.CertificateIssuer([
-                        cryptography_get_name(name, 'issuer') for name in entry['issuer']
```
felixfontein, line 693: `entry['issuer']` is already transformed in line 530.

tests/integration/targets/x509_crl/tests/validate.yml
```diff
@@ -102,3 +102,11 @@
...
           ['commonName', 'Test'],
         ]
+
+- name: Validate CRL 3 info
+  assert:
+    that:
+      - crl_3.revoked_certificates == crl_3_info.revoked_certificates
+      - crl_3.revoked_certificates[0].issuer == [
+          "DNS:ca.example.org",
+        ]
```
briantist, line 112: I was surprised that this plain scalar worked multi-line without `>` or `>-` or something... I did find at least one yaml parser online that didn't like it: https://codebeautify.org/yaml-validator/y22bd59f9

But it appears that validator is wrong; and it clearly works in Ansible regardless, but you may want to change it if only for consistency with the other asserts in the file.

Or not 🤷

briantist (approved):

tests/integration/targets/x509_crl/tests/validate.yml
felixfontein: According to the YAML definition this should be fine: https://yaml.org/spec/1.2.2/#plain-style
I'd probably rather remove the `>` from the other blocks in this file (but not in this PR) :)

felixfontein:
@briantist thanks for reviewing this!

PR #448, opened 2022-04-29
Make request timeout configurable for all acme modules

### LINKED ISSUE #447: Acme_* modules should have configurable timeouts
##### SUMMARY
As far as I can tell, all `acme_`-modules use the standard `fetch_url` timeout, [which is 10 seconds](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py#L1729). (They all call the [acme-util](https://github.com/ansible-collections/community.crypto/blob/main/plugins/module_utils/acme/acme.py))
This timeout should be configurable in case of a slow responding ACME server.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
acme_account
acme_account_info
acme_certificate
acme_certificate_revoke
acme_challenge_cert_helper
acme_inspect

##### ADDITIONAL INFORMATION
I'm not entirely sure if this should be a global variable somehow or should be set at the module level on a per task-basis?
I've hacked our local ansible instance and have confirmed that changing the default to something higher makes 95% of our requests succeed again. (many were failing because our cert provider is having issues..)

Would gladly whip up a PR, but I'm unsure what the best location would be to set the timeout.

JonasVerhofste:
##### SUMMARY
Makes request timeouts configurable for all `acme_`-modules that use the shared acme argspec. Fixes #447.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`plugins/module_utils/acme/acme.py`; affected modules:
  acme_account.py
  acme_account_info.py
  acme_inspect.py
  acme_certificate.py
  acme_certificate_revoke.py

##### ADDITIONAL INFORMATION
Added `timeout` as an alias because it seemed logical, not sure if that's allowed (i.e. if that is usually only used for backwards compatible changes).

### DIFF
--- a/changelogs/fragments/448-acme-request-timeouts.yml
+++ b/changelogs/fragments/448-acme-request-timeouts.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - acme_* modules - add parameter ``request_timeout`` to manage HTTP(S) request timeout (https://github.com/ansible-collections/community.crypto/issues/447, https://github.com/ansible-collections/community.crypto/pull/448).
--- a/plugins/doc_fragments/acme.py
+++ b/plugins/doc_fragments/acme.py
@@ -123,4 +123,11 @@ options:
     type: str
     default: auto
     choices: [ auto, cryptography, openssl ]
+  request_timeout:
+    description:
+      - The time Ansible should wait for a response from the ACME API.
+      - This timeout is applied to all HTTP(S) requests (HEAD, GET, POST).
+    type: int
+    default: 10
+    version_added: 2.3.0
 '''
--- a/plugins/module_utils/acme/acme.py
+++ b/plugins/module_utils/acme/acme.py
@@ -83,6 +83,8 @@ class ACMEDirectory(object):
 
         self.directory, dummy = account.get_request(self.directory_root, get_only=True)
 
+        self.request_timeout = module.params['request_timeout']
+
         # Check whether self.version matches what we expect
         if self.version == 1:
             for key in ('new-reg', 'new-authz', 'new-cert'):
@@ -103,7 +105,7 @@ class ACMEDirectory(object):
         url = self.directory_root if self.version == 1 else self.directory['newNonce']
         if resource is not None:
             url = resource
-        dummy, info = fetch_url(self.module, url, method='HEAD')
+        dummy, info = fetch_url(self.module, url, method='HEAD', timeout=self.request_timeout)
         if info['status'] not in (200, 204):
             raise NetworkException("Failed to get replay-nonce, got status {0}".format(info['status']))
         return info['replay-nonce']
@@ -130,6 +132,8 @@ class ACMEClient(object):
         # Grab account URI from module parameters.
         # Make sure empty string is treated as None.
         self.account_uri = module.params.get('account_uri') or None
+
+        self.request_timeout = module.params['request_timeout']
 
         self.account_key_data = None
         self.account_jwk = None
@@ -235,7 +239,7 @@ class ACMEClient(object):
             headers = {
                 'Content-Type': 'application/jose+json',
             }
-            resp, info = fetch_url(self.module, url, data=data, headers=headers, method='POST')
+            resp, info = fetch_url(self.module, url, data=data, headers=headers, method='POST', timeout=self.request_timeout)
             _assert_fetch_url_success(self.module, resp, info)
             result = {}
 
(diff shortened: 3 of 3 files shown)


JonasVerhofste:
Oh right, let me add some tests..

JonasVerhofste:
@felixfontein Where would you suggest adding this to the tests (if needed at all)?

plugins/doc_fragments/acme.py
```diff
@@ -123,4 +123,11 @@ class ModuleDocFragment(object):
     type: str
     default: auto
     choices: [ auto, cryptography, openssl ]
+  request_timeout:
+    description:
+      - The time ansible should wait for a response from the ACME API.
+      - This timeout is applied to all HTTP(S) requests (HEAD, GET, POST).
+    type: int
+    default: 10
+    aliases: [ timeout ]
```
felixfontein, line 132:
```suggestion
    aliases: [ timeout ]
    version_added: 2.3.0
```

plugins/doc_fragments/acme.py
```diff
@@ -123,4 +123,11 @@ class ModuleDocFragment(object):
     type: str
     default: auto
     choices: [ auto, cryptography, openssl ]
+  request_timeout:
+    description:
+      - The time ansible should wait for a response from the ACME API.
```
felixfontein, line 128:
```suggestion
      - The time Ansible should wait for a response from the ACME API.
```

felixfontein (commented):
Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)?

About the alias: I would not add `timeout` as an alias to make it possible to use that name for something else in the future. For example, `acme_certificate` could have a `timeout` for the second stage where it eventually gives up waiting for the finalization to take place. Once such a functionality is there, I would expect `timeout` to be in control of that one. (It will probably be best to use a different name for that one as well, and not use `timeout` at all, to make it less confusing.)

felixfontein:
> @felixfontein Where would you suggest adding this to the tests (if needed at all)?

Timeouts are hard to test with integration tests (without having an endpoint supporting that usage), so unit tests would be best. But since we right now have to unit tests for anything HTTP related, I think it's ok to not add tests here.

JonasVerhofste:
@felixfontein removed the `timeout` alias and added the changelog fragment!

changelogs/fragments/448-acme-request-timeouts.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - acme - Add parameter to manage HTTP(S) request timeout.
```
felixfontein, line 3:
```suggestion
  - acme_* modules - add parameter ``request_timeout`` to manage HTTP(S) request timeout (https://github.com/ansible-collections/community.crypto/issues/447, https://github.com/ansible-collections/community.crypto/pull/448).
```

JonasVerhofste: Done!

changelogs/fragments/448-acme-request-timeouts.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - acme_* modules - Add parameter ``request_timeout`` to manage HTTP(S) request timeout (https://github.com/ansible-collections/community.crypto/issues/447, https://github.com/ansible-collections/community.crypto/pull/448).
```
felixfontein, line 3: Just a tiny thing that got missing in the last change :)
```suggestion
  - acme_* modules - add parameter ``request_timeout`` to manage HTTP(S) request timeout (https://github.com/ansible-collections/community.crypto/issues/447, https://github.com/ansible-collections/community.crypto/pull/448).
```

JonasVerhofste: Ah, I thought that was a typo :sweat_smile: 
Sentences start with a Capital and end with a punctuation mark =P

Anyhow, lowercased it :)

felixfontein: Sentences do, but these are not proper sentences :) I don't know who came up with this, but since it was the convention in ansible/ansible for quite some time I'm trying to make sure that all fragments look the same way. Makes the changelog easier to read if such things don't change from entry to entry ;-) (https://docs.ansible.com/ansible/devel/community/collection_development_process.html#changelog-fragment-entry-format)

felixfontein (approved):

felixfontein:
@JonasVerhofste thanks for implementing this!

PR #452, opened 2022-05-09
openssl_pkcs12: add privatekey_content option

felixfontein:
##### SUMMARY
This allows to store the private key in another form, like in Ansible vault, or encrypted with sops.

Right now it has to be present as a file, which in my setup requires me to write it to a temporary file (unprotected), use it in this module, and then wipe the temporary file in a `always:` part of a `block:`. That's a lot of extra code and also dangerous.

(Not that PKCS12 passphrases offer *that* much more protection...)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
openssl_pkcs12

### DIFF
--- a/changelogs/fragments/452-openssl_pkcs12-private-key-content.yml
+++ b/changelogs/fragments/452-openssl_pkcs12-private-key-content.yml
@@ -1,0 +1,4 @@
+minor_changes:
+  - "openssl_pkcs12 - allow to provide the private key as text instead of having to read it from a file.
+     This allows to store the private key in an encrypted form, for example in Ansible Vault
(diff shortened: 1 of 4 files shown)


s-hertel (approved):

briantist (approved):
Yes! Love to see changes like this. HashiCorp Vault could be another source for the key, just sayin' 😉

felixfontein:
@s-hertel @briantist thanks a lot for reviewing this!

HashiCorp vault is another use-case for this, as is sops. The latter triggered me to do this, since I use sops for my private infrastructure and had to resort to this hack ;-)

PR #457, opened 2022-05-19
certificate_complete_chain: do not stop execution on unsupported algorithm

felixfontein:
##### SUMMARY
Found while working on #456.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
certificate_complete_chain

PR #408, opened 2022-02-19
Integrate Alpine into CI

felixfontein:
##### SUMMARY
Let's see how smooth this one goes.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

PR #463, opened 2022-05-20
ci: enable rhel9.0 tests for openssh_cert

### LINKED ISSUE #462: openssh_cert: make tests work with RHEL 9.0
##### SUMMARY
They are currently disabled (#456); ssh-keygen crashed with an error message.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
openssh_cert

Ajpantuso:
##### SUMMARY

Re-enables RHEL9.0 tests for `openssh_cert`
Fixes #462

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/tests/integration/targets/openssh_cert/tests/key_idempotency.yml
+++ b/tests/integration/targets/openssh_cert/tests/key_idempotency.yml
@@ -63,13 +63,13 @@
         valid_from: always
         valid_to: forever
       register: second_signature_algorithm
-      when: openssh_version is version("8.7", "<")
+      when: openssh_version is version("8.7", "!=")
 
     - name: Assert second signature algorithm update causes change
       assert:
         that:
           - second_signature_algorithm is changed
-      when: openssh_version is version("8.7", "<")
+      when: openssh_version is version("8.7", "!=")
 
     - name: Omit signature algorithm
       openssh_cert:
--- a/tests/integration/targets/openssh_cert/tests/key_idempotency.yml
+++ b/tests/integration/targets/openssh_cert/tests/key_idempotency.yml
@@ -63,11 +63,13 @@
         valid_from: always
         valid_to: forever
       register: second_signature_algorithm
+      when: openssh_version is version("8.7", "<")
 
     - name: Assert second signature algorithm update causes change
       assert:
         that:
           - second_signature_algorithm is changed
+      when: openssh_version is version("8.7", "<")
 
     - name: Omit signature algorithm
       openssh_cert:
--- a/tests/integration/targets/openssh_cert/tests/key_idempotency.yml
+++ b/tests/integration/targets/openssh_cert/tests/key_idempotency.yml
@@ -63,13 +63,13 @@
         valid_from: always
         valid_to: forever
       register: second_signature_algorithm
-      when: openssh_version is version("8.7", "!=")
+      when: not (ansible_facts['distribution'] == "RedHat" and ansible_facts['distribution_major_version'] >= "9")
 
     - name: Assert second signature algorithm update causes change
       assert:
         that:
           - second_signature_algorithm is changed
-      when: openssh_version is version("8.7", "!=")
+      when: not (ansible_facts['distribution'] == "RedHat" and ansible_facts['distribution_major_version'] >= "9")
 
     - name: Omit signature algorithm
       openssh_cert:
--- a/tests/integration/targets/openssh_cert/tests/key_idempotency.yml
+++ b/tests/integration/targets/openssh_cert/tests/key_idempotency.yml
@@ -53,23 +53,25 @@
         that:
           - updated_signature_algorithm_idempotent is not changed
 
-    - name: Generate cert with original signature algorithm
-      openssh_cert:
-        type: user
-        path: "{{ certificate_path }}"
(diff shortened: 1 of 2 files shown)


Ajpantuso:
Not really confident on the incompatibility introduced with _OpenSSH 8.7_. It could be related to the underlying _OpenSSL_ version or an unintended consequence of staging the `RSA1` deprecation, but either way signing keys with the old algorithm will be fully unsupported in the future.

MarkusTeufelberger:
Maybe we should test on something more bleeding edge like Arch or Alpine too?

Ajpantuso:
> Maybe we should test on something more bleeding edge like Arch or Alpine too?

We do test with Alpine (See #408) and these [tests are passing](https://dev.azure.com/ansible/community.crypto/_build/results?buildId=42952&view=logs&j=703a0d87-7854-5908-28c0-1801c01335e4&t=a73dccc4-d146-506c-e684-ded38204ddd2&l=2924) with _Openssh 8.8_ which indicates this was an ephemeral bug. I'll pin the skip condition to 8.7 and see what happens.

felixfontein:
FYI: we also run tests with Arch: https://github.com/ansible-collections/community.crypto/blob/main/.azure-pipelines/azure-pipelines.yml#L217 - and the tests don't fail there.

RHEL 9 seems to have disabled SHA1 completely, though - maybe it's even FIPS mode? I haven't checked that. That was causing trouble with the `certificate_complete_chain` tests (it failed when tring to verify a SHA1 based signature on a root certificate; that's also why we got #457). Maybe this is related?

tests/integration/targets/openssh_cert/tests/key_idempotency.yml
```diff
@@ -63,11 +63,13 @@
         valid_from: always
         valid_to: forever
       register: second_signature_algorithm
+      when: openssh_version is version("8.7", "!=")
```
felixfontein, line 66: This is likely the wrong check, since the tests pass with OpenSSH 9.0 on Arch Linux: https://dev.azure.com/ansible/community.crypto/_build/results?buildId=42954&view=logs&j=57f52a93-2f8f-5f4a-64e7-e08c30d14461&t=639a25f5-5b1b-5d56-c79d-060a9682ed36&l=312

felixfontein: How about checking for RHEL 9 instead?

felixfontein: Another datapoint: the tests pass with OpenSSH 8.7 on Fedora 35: https://dev.azure.com/ansible/community.crypto/_build/results?buildId=42954&view=logs&j=ad47e826-3d51-5052-c5d1-abf94a7b75a5&t=a7141e3f-34f2-5aff-b04a-ec77c032fc27&l=351

So it's really not 8.7 in general, but something that RHEL 9 does.

Ajpantuso: Ah, not sure if you all have access to [this](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/considerations_in_adopting_rhel_9/assembly_security_considerations-in-adopting-rhel-9#ref_considerations-security-crypto_changes-to-security), but RHEL9 disables SHA1 in the default crypto-policies and would have to be explicitly enabled. I'll push the updated condition in a moment.

tests/integration/targets/openssh_cert/tests/key_idempotency.yml
```diff
@@ -63,13 +63,13 @@
...
       register: second_signature_algorithm
-      when: openssh_version is version("8.7", "!=")
+      when: not (ansible_facts['distribution'] == "RedHat" and ansible_facts['distribution_major_version'] >= "9")
 
     - name: Assert second signature algorithm update causes change
       assert:
         that:
           - second_signature_algorithm is changed
-      when: openssh_version is version("8.7", "!=")
+      when: not (ansible_facts['distribution'] == "RedHat" and ansible_facts['distribution_major_version'] >= "9")
```
felixfontein, line 72: Otherwise `"10"` won't be `>= "9"` :-)
```suggestion
      when: not (ansible_facts['distribution'] == "RedHat" and (ansible_facts['distribution_major_version'] | int) >= 9)
```
Also how about adding the `when` to ` block` containing both tasks, and adding a comment that this `when` is needed for systems not allowing SHA1? (There could be more such systems in the future.)

tests/integration/targets/openssh_cert/tests/key_idempotency.yml
felixfontein: Thanks! That document seems to be publicly visible.

tests/integration/targets/openssh_cert/tests/key_idempotency.yml
Ajpantuso: Done

felixfontein (approved):
Looks good, thanks!

felixfontein:
@Ajpantuso thanks for fixing this!
@MarkusTeufelberger thanks for reviewing!

PR #478, opened 2022-06-15
Add Apache 2.0 license; simplify and standardize license headers

felixfontein:
##### SUMMARY
Some vendored data from OpenSSL and code from cryptography is Apache 2.0 licensed, but we didn't include the license file.

Also moves the additional licenses to licenses/, and unifies the license headers in source files.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
licenses


plugins/module_utils/crypto/cryptography_crl.py
```diff
@@ -1,19 +1,7 @@
...
-# (at your option) any later version.
-#
-# Ansible is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
gotmax23, line 4: It would be helpful if you could label files with a comment such as `# SPDX-License-Identifier: GPL-3.0-or-later` or `# SPDX-License-Identifier: BSD-2-Clause`. License scanners can detect the previous license header as `GPL-3.0-or-later`, but not the new one (at least not consistently), as it doesn't follow any standardized format.

I would also recommend looking into the [REUSE.software](https://REUSE.software) standard. KDE and the Linux kernel, amongst others, follow this, at least to some extent.

plugins/module_utils/crypto/_obj2txt.py
```diff
@@ -1,7 +1,15 @@
+# This code is part of Ansible, but is an independent component.
+# This particular file snippet, and this file snippet only, is licensed under the
+# Apache 2.0 License. Modules you write using this snippet, which is embedded
+# dynamically by Ansible, still belong to the author of the module, and may assign
+# their own license to the complete work.
+
 # This excerpt is dual licensed under the terms of the Apache License, Version
```
gotmax23, line 7: For example, the proper SPDX license identifier for this file is `# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause`.

plugins/module_utils/crypto/cryptography_crl.py
gotmax23: Even if you don't want to add an `SPDX-License-Identifier` comment to every file, like REUSE dictates, you could start by changing the name of the `licenses` directory to `LICENSES` and change `PSF-license.txt` > `PSF-2.0.txt` and `simplified_bsd.txt` > `BSD-2-Clause.txt`.

felixfontein: Hmm, this simplified header is standard in ansible-core and a lot of other collections (like community.general). I don't like the idea of replacing it (or the copyright statements above it) completely with the notation in https://reuse.software/spec/, especially since the current simplified version is a lot easier to read for users. But I guess it would be OK to add the `SPDX-License-Identifier` line directly after this simplified header.

Also the `licenses` directory and the filenames in there are as in ansible-core (https://github.com/ansible/ansible/tree/devel/licenses), and the main project license is in `COPYING` outside this directory.

felixfontein:
I reverted the move to `licenses/` for now, I think this should be discussed more first.

Whether to add `SPDX-License-Identifier` should also be discussed, but shouldn't hold up this PR either.

Andersson007 (approved):

plugins/module_utils/crypto/cryptography_crl.py
felixfontein: I created https://github.com/ansible-community/community-topics/issues/112 to discuss this.

felixfontein:
@gotmax23 @Andersson007 thanks a lot for reviewing and starting the discussion!

Andersson007:
@felixfontein thanks for working on this!

gotmax23:
No problem! I will chime in on the community-topics discussion when I get a chance. Thanks for your work on this.

PR #491, opened 2022-07-17
Move licenses to LICENSES/, use SPDX-License-Identifier, mention all licenses in galaxy.yml

felixfontein:
##### SUMMARY
Implements https://github.com/ansible-community/community-topics/issues/112#issuecomment-1166348220.

##### ISSUE TYPE
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME
collection


felixfontein:
I created https://github.com/ansible/galaxy-importer/pull/175 to address the failing galaxy import.

briantist (approved):

plugins/module_utils/crypto/_obj2txt.py
```diff
@@ -8,7 +8,8 @@
 # 2.0, and the BSD License. See the LICENSE file at
 # https://github.com/pyca/cryptography/blob/master/LICENSE for complete details.
```
felixfontein, line 9: @gotmax23 this file is dually licensed in its original source, but I chose to only use one of the licenses (Apache 2.0) here since the other (BSD 3-clause, https://github.com/pyca/cryptography/blob/main/LICENSE.BSD) would be yet another license to list that isn't used anywhere else in this collection. Is it OK to only add `SPDX-License-Identifier` for `Apache-2.0` in that case?

gotmax23: In order to comply with the license terms, you don't *have* to, but I would recommend doing so for two reasons:
1. It's more accurate. The code is licensed under either of the licenses, not just one of them. This is how we handle it in Fedora when determining the License tag for a package, as well. Even if it seems obvious to use one choice, because that makes it so less licenses apply overall, Fedora Legal recommends including both.
2. If you only license this under one of the licenses and then make changes, it would be impossible to submit them upstream.

felixfontein: Hmm, but that would mean that we would have to add yet another license to this collection (one more file, one more entry in galaxy.yml), which sounds very excessive.

gotmax23: > Hmm, but that would mean that we would have to add yet another license to this collection (one more file, one more entry in galaxy.yml), which sounds very excessive.

I don't deny that, but I maintain my recommendation. Not my call though :).

briantist: Personally I think if the license applies, then include it (both in `LICENSES/` and in `galaxy.yml`). Part of what's nice about having the directory for licenses is not littering them in the root, and in `galaxy.yml` it's just another list entry.

felixfontein:
@gotmax23 @briantist I've added BSD-3-Clause (with the license text from cryptography, which differs in two locations from the one on https://spdx.org/licenses/BSD-3-Clause: `PyCA Cryptography` instead of `copyright holder` in clause 3, and `COPYRIGHT OWNER` instead of `COPYRIGHT HOLDER` in the all-caps part), and also updated the README to mention all licenses.

felixfontein:
CI errors are unrelated.

felixfontein:
@briantist @gotmax23 any last words before merging this?

briantist (approved):

PR #495, opened 2022-07-24
Correctly mark plugins/module_utils/crypto/_obj2txt.py as being dually licensed

felixfontein:
##### SUMMARY
I think I marked that file wrong. (See https://reuse.software/spec/#comment-headers)

CC @gotmax23 @briantist

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
plugins/module_utils/crypto/_obj2txt.py

### DIFF
--- a/plugins/module_utils/crypto/_obj2txt.py
+++ b/plugins/module_utils/crypto/_obj2txt.py
@@ -10,8 +10,7 @@
 #
 # The Apache 2.0 license has been included as LICENSES/Apache-2.0.txt in this collection.
 # The BSD License license has been included as LICENSES/BSD-3-Clause.txt in this collection.
-# SPDX-License-Identifier: Apache-2.0
-# SPDX-License-Identifier: BSD-3-Clause
+# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 #
 # Adapted from cryptography's hazmat/backends/openssl/decode_asn1.py
 #
--- a/tests/sanity/extra/licenses.py
+++ b/tests/sanity/extra/licenses.py
@@ -30,7 +30,8 @@ def find_licenses(filename, relax=False):
                 print('%s: found copyright line with "Copyright:". Please remove the colon.' % (filename, ))
             idx = line.find('SPDX-License-Identifier: ')
             if idx >= 0:
-                spdx_license_identifiers.append(line[idx + len('SPDX-License-Identifier: '):])
+                lic_id = line[idx + len('SPDX-License-Identifier: '):]
+                spdx_license_identifiers.extend(lic_id.split(' OR '))
             if 'GNU General Public License' in line:
                 if 'v3.0+' in line:
                     other_license_identifiers.append('GPL-3.0-or-later')


gotmax23:
That's correct. If a file is dual licensed, the two licenses should be separated with `OR`. If multiple licenses apply to a file at the same time, they should be separated with `AND`.

It also allows grouping. The SPDX license identifier of this whole collection would be something like `GPL-3.0-or-later AND Apache-2.0 AND BSD-2-Clause AND PSF-2.0 AND (Apache-2.0 OR BSD-3-Cluase)`.

tests/sanity/extra/licenses.py
```diff
@@ -30,7 +30,8 @@ def find_licenses(filename, relax=False):
                 print('%s: found copyright line with "Copyright:". Please remove the colon.' % (filename, ))
             idx = line.find('SPDX-License-Identifier: ')
             if idx >= 0:
-                spdx_license_identifiers.append(line[idx + len('SPDX-License-Identifier: '):])
+                lic_id = line[idx + len('SPDX-License-Identifier: '):]
+                spdx_license_identifiers.extend(lic_id.split(' OR '))
```
gotmax23, line 34: This should at least account for identifiers split with `AND` if not also license identifiers that have groupings with multiple boolean conditions. It would obviously require more code, but future proofing for if collections do start using this is a good idea in my book.

plugins/module_utils/crypto/_obj2txt.py
```diff
@@ -10,8 +10,7 @@
 #
 # The Apache 2.0 license has been included as LICENSES/Apache-2.0.txt in this collection.
 # The BSD License license has been included as LICENSES/BSD-3-Clause.txt in this collection.
-# SPDX-License-Identifier: Apache-2.0
-# SPDX-License-Identifier: BSD-3-Clause
+# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
```
gotmax23, line 13: This looks good and matches upstream's license text:

> This software is made available under the terms of *either* of the licenses
found in LICENSE.APACHE or LICENSE.BSD. Contributions to cryptography are made
under the terms of *both* these licenses.

tests/sanity/extra/licenses.py
gotmax23: If you implement grouping, you'll also want to remove the `found identical SPDX-License-Identifier values` check.

gotmax23 (commented):
The license part is correct (good catch!), but I had some comments about the linting script changes. I'm fine with merging this as is and then iterating on the script later if desired.

tests/sanity/extra/licenses.py
felixfontein: I think I only want to implement the bare minimum to support what's needed in this collection. The default for files should be to have exactly one license, with any other combination (like `OR`) reserved for very special circumstances.

gotmax23 (approved):

felixfontein:
@gotmax23 thanks a lot for reviewing this!

gotmax23:
Sure!

PR #502, opened 2022-08-21
Make mostly reuse conformant

felixfontein:
##### SUMMARY
Everything except changelog/fragments/ is now reuse conformant (https://reuse.software/spec/).

CC @gotmax23

##### ISSUE TYPE
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME
licensing


felixfontein:
@gotmax23 one thing I'm *very* unsure about is the last commit: 35e106867c026393e144a2932239e5742bb3fb18

It adds `.license` files for all the vendored PEM certificate files that are used in some of the tests. I'm not sure whether these files are actually copyrightable / licensable, and if yes which license / copyright text should be used. Right now I put the default collections' license + copyright combination in there, but it feels wrong.

Maybe you saw something similar and have an idea what to do here?

gotmax23:
> @gotmax23 one thing I'm _very_ unsure about is the last commit: [35e1068](https://github.com/ansible-collections/community.crypto/commit/35e106867c026393e144a2932239e5742bb3fb18)
> 
> It adds `.license` files for all the vendored PEM certificate files that are used in some of the tests. I'm not sure whether these files are actually copyrightable / licensable, and if yes which license / copyright text should be used. Right now I put the default collections' license + copyright combination in there, but it feels wrong.
> 
> Maybe you saw something similar and have an idea what to do here?

IANAL, and I'm not exactly sure what to do here. I guess you have three options:

- Leave it as is with the explicit GPL-3.0-or-later license. There was already an implicit GPL license.
- You can use `SPDX-License-Identifier: Unlicense` if you want to declare it as non-copyrightable. REUSE recommends `CC0` for this purpose, but the RH lawyers don't like that license and they've recently banned it in Fedora.
- Not declare any explicit license and break REUSE compliance.

felixfontein:
I'm currently tending to the third option. The first two require legal advice (IMO)... I guess a better approach is to replace them by files we generate ourselves. This will take a bit of time though, but I think I'll work on that first.

gotmax23:
> I'm currently tending to the third option. The first two require legal advice (IMO)

I'd lean towards adding the GPL headers (option 1). That's what was already there, even if it was implicit, so we wouldn't be changing anything. If the files aren't copyrightable, then the license is invalid, anyways. It's no different than the blank `__init__.py` files that we added headers to.

FWIW, I do remember that one of the go libraries used by docker that we package in Fedora includes a Lets Encrypt root certificate that it uses for tests, as well. The library is Apache licensed. There was no special license information attached to it. The certificate is copied straight into a `const` in the go source file. (I know about this, because I had to update the certificate to a non-expired one to fix test failures)

tests/sanity/extra/licenses.py
```diff
@@ -21,30 +21,35 @@ def find_licenses(filename, relax=False):
...
+                if 'Simplified BSD License' in line:
+                    other_license_identifiers.append('BSD-2-Clause')
+                if 'Apache License 2.0' in line:
+                    other_license_identifiers.append('Apache-2.0')
+                if 'PSF License' in line or 'Python-2.0' in line:
+                    other_license_identifiers.append('PSF-2.0')
+                if 'MIT License' in line:
+                    other_license_identifiers.append('MIT')
+    except Exception as exc:
+        print('%s: error while processing file: %s' % (filename, exc))
```
gotmax23, line 52: So the only change here is putting it in a try-except block?

gotmax23 (approved):

tests/sanity/extra/licenses.py
felixfontein: And support for `SPDX-FileCopyrightText:`. The try/except block comes from community.general, the `SPDX-FileCopyrightText:` block change too I think, and the other projects got the `OR` change from this one IIRC. Basically making sure that all projects have essentially the same copy of this whole file. (The only exception is community.dns, which has a special clause for `plugins/public_suffix_list.dat` having a non-GPL license outside of module_utils and modules.)

felixfontein:
I decided to revert the controversial part of this PR so we can get it merged, and only have the controversial part left.

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: |
+          pip install reuse
+
+      - name: Check REUSE compliance (currently optional)
+        run: |
+          reuse lint || true
```
gotmax23, line 32: Why not `rm` the files we're skipping like we previously did for the changelog fragments?

felixfontein: Good point, improved that in 7eddfda7f850d17a6f90862ce061a74d1fb364f7.

felixfontein:
@gotmax23 thanks for reviewing this, and your comments!

felixfontein:
The controversial parts and reverts of the last commits are in #509.

gotmax23:
> @gotmax23 thanks for reviewing this, and your comments!

Sure! Thank *you* for all of the work you've done for this change.

PR #509, opened 2022-09-13
Make reuse conformant

felixfontein:
##### SUMMARY
Reverts the controversial commit from #502 and all the other changes added to make the new tests pass.

##### ISSUE TYPE
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME
licensing


felixfontein:
Hmm, I found this on https://www.mono-project.com/docs/faq/security/:

> Digital certificates are, like many digital files, copyrightable.

They use this as one reason why they don't include root certificates.

TBH, this looks like something that needs a decision in court until we really know whether such files are copyrightable or not.

Maybe it's best to try to replace them with self-generated files, if possible.

felixfontein:
#862 replaced the problematic certificates, so now we're definitely on the safe side. :tada:

Issue #518, opened 2022-10-13
Can't detect cryptography even though its installed


Subito:
##### SUMMARY

Running `acme_certificate`, I get the following error:
```
fatal: [remote_host]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (cryptography) on remote_host's Python /usr/local/bin/python. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"}
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

acme_certificate
##### ANSIBLE VERSION

```paste below
ansible 2.10.17
  config file = /Users/svbito/src/nkhosting/ansible/ansible.cfg
  configured module search path = ['/Users/svbito/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/svbito/.virtualenvs/ansible29/lib/python3.9/site-packages/ansible
  executable location = /Users/svbito/.virtualenvs/ansible29/bin/ansible
  python version = 3.9.15 (main, Oct 11 2022, 22:27:25) [Clang 14.0.0 (clang-1400.0.29.102)]
```

##### COLLECTION VERSION

```paste below
# /Users/svbito/.ansible/collections/ansible_collections
Collection       Version
---------------- -------
community.crypto 2.7.0
```

##### OS / ENVIRONMENT

Target OS: FreeBSD 13.1

##### STEPS TO REPRODUCE

```yaml
- name: Let the challenge be validated and retrieve the cert and intermediate certificate
  community.crypto.acme_certificate:
    account_key_src: /usr/local/etc/ssl/letsencrypt-account.key
    account_email: someone@example.com
    src: /usr/local/etc/ssl/{{domain}}.csr
    cert: /usr/local/etc/ssl/{{domain}}.crt
    fullchain: /usr/local/etc/ssl/{{domain}}_fullchain.crt
    chain: /usr/local/etc/ssl/{{domain}}_chain.crt
    challenge: dns-01
    acme_directory: https://acme-v02.api.letsencrypt.org/directory
    remaining_days: 30
    acme_version: 2
    data: "{{ challenge }}"
    select_crypto_backend: cryptography
    select_chain:
      - test_certificates: last
        issuer:
          CN: ISRG Root X1
```

##### EXPECTED RESULTS

The cert gets fetched

##### ACTUAL RESULTS

```
[...]

PR #519, opened 2022-10-13
Improve import error handling for ACME modules

felixfontein:
##### SUMMARY
Improve cryptography import / prepare error handling.

Ref #518.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
acme module utils

### DIFF
--- a/plugins/module_utils/acme/acme.py
+++ b/plugins/module_utils/acme/acme.py
@@ -27,6 +27,8 @@ from ansible_collections.community.crypto.plugins.module_utils.acme.backend_open
 
 from ansible_collections.community.crypto.plugins.module_utils.acme.backend_cryptography import (
     CryptographyBackend,
+    CRYPTOGRAPHY_ERROR,
+    CRYPTOGRAPHY_MINIMAL_VERSION,
     CRYPTOGRAPHY_VERSION,
     HAS_CURRENT_CRYPTOGRAPHY,
 )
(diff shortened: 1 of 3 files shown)


plugins/module_utils/acme/acme.py
```diff
@@ -399,8 +401,19 @@ def create_backend(module, needs_acme_v2):
 
     # Create backend object
     if backend == 'cryptography':
+        if CRYPTOGRAPHY_ERROR is not None:
+            # Either we couldn't import cryptography at all, or there was an unexpected error
+            if CRYPTOGRAPHY_VERSION is None:
+                msg = missing_required_lib('cryptography')
+            else:
+                msg = 'Unexpected error while preparing cryptography: {0}'.format(CRYPTOGRAPHY_ERROR.splitlines()[-1])
+            module.fail_json(msg=msg, exception=CRYPTOGRAPHY_ERROR)
```
webknjaz, line 410: Is the argument name `exception`? I thought it's `traceback`.

plugins/module_utils/acme/acme.py
```diff
@@ -399,8 +401,19 @@ def create_backend(module, needs_acme_v2):
...
         if not HAS_CURRENT_CRYPTOGRAPHY:
-            module.fail_json(msg=missing_required_lib('cryptography'))
+            # We could import cryptography, but an too old version.
```
webknjaz, line 412:
```suggestion
            # We could import cryptography, but the version might be too old.
```

felixfontein: At this point we know that the version is too old, so 'might be' doesn't sound correct to me. How about:
```suggestion
            # We could import cryptography, but the version is too old.
```
?

felixfontein: or even better
```suggestion
            # We can import cryptography, but the version is too old.
```

webknjaz:
```suggestion
            # We are able to import cryptography, but the version is too old.
```

webknjaz:
```suggestion
            # We succeeded importing cryptography, but its version is too old.
```

felixfontein:
@webknjaz thanks for reviewing this!

PR #520, opened 2022-10-24
CI: drop remote CI targets for EoL Ansible/ansible-base versions; move docker CI targets for EoL Ansible/ansible-base versions to GHA

felixfontein:
##### SUMMARY
Drop remote CI targets for EoL Ansible/ansible-base versions.
Move docker CI targets for EoL Ansible/ansible-base versions to GHA.

This should remove the size of the Docker image collection on AZP.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -109,28 +109,6 @@ stages:
               test: '2.11/sanity/1'
             - name: Units
               test: '2.11/units/1'
-  - stage: Ansible_2_10
-    displayName: Sanity & Units 2.10
-    dependsOn: []
-    jobs:
-      - template: templates/matrix.yml
-        parameters:
-          targets:
-            - name: Sanity
-              test: '2.10/sanity/1'
-            - name: Units
-              test: '2.10/units/1'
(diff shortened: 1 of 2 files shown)


felixfontein:
This isn't perfect (since there's no change detection), but at least it improves AZP usage.

felixfontein:
https://github.com/ansible-community/ansible-test-gh-action/pull/46 adds support for change detection to the action.

It would also be useful to not always run all tests (`--continue-on-error` option) for integration tests. While that can be useful if coverage is computed, it makes the test take a lot longer than necessary.

felixfontein:
(Always running all tests also makes it **a lot harder** to find actual errors. In https://github.com/ansible-collections/community.crypto/actions/runs/3314925069/jobs/5475617923#step:3:21286 I had to scroll down thousands of lines to find the error in line 21286 out of 49914. Would be *a lot* easier if the error would just be at the bottom, as with AZP. Sometimes even then it's hard enough to find it if it is hidden by tons of pages shutting down the test environment...)

felixfontein:
I'm going to merge this PR as-is so we can try this out a bit...

PR #530, opened 2022-11-17
fix #529 issuer_uri in x509_certificate_info

### LINKED ISSUE #529: Retrieve Authority Information Access via x509_certificate_info
##### SUMMARY

Improve *community.crypto.x509_certificate_info* to also yield the contents of *Authority Information Access* (OID 1.3.6.1.5.5.7.1.1).

Alternatively:

- allow custom OIDs to be ASN.1 parsed by the module or
- have a filter for parsing the current output (b64'd ASN.1) to usable values within Ansible (or provide documentation on how to do so if already possible)

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`x509_certificate_info`

##### ADDITIONAL INFORMATION

AIA is present in most if not all certificates, and it allows one to piece together a certificate chain based on a single certificate.
This is incredibly useful when only one certificate or an invalid chain is present, since it allows one to recreate a valid chain in most cases.

```yaml
- name: Get certificate information
  community.crypto.x509_certificate_info:
    path: /some/certificate.pem
  register: certinfo

- name: Get upstream certificate
  ansible.builtin.debug:
    # currently the next best thing is to read the appropriate OID, which yields base64'd ASN.1 which is hard to parse in Ansible directly
    # one *could* pass the result off to another task to fetch the certificate then
    # this prints "MEcwIQYIKwYBBQUHMAGGFWh0dHA6Ly9yMy5vLmxlbmNyLm9yZzAiBggrBgEFBQcwAoYWaHR0cDovL3IzLmkubGVuY3Iub3JnLw==" for a Let's Encrypt cert
    msg: |-
      {{ certinfo.extensions_by_oid["1.3.6.1.5.5.7.1.1"] }}

- name: Get upstream certificate (decoded)
  ansible.builtin.debug:
    # currently even base64 decoding the certificate causes issues in Ansible since either Ansible or Jinja don't seem to work well with binary values
    # this should in theory yield "lencr" from the "http://r3.i.lencr.org/" link contained, but it does yield an empty string
    msg: |-
      {{ certinfo.extensions_by_oid["1.3.6.1.5.5.7.1.1"] | b64decode | regex_search("[a-z]{5}", multiline=True) }}
```

Basically I'm looking for any means (short of writing a custom plugin/filter/module) to retrieve the `http://r3.i.lencr.org/` string from the cert.
Note that this is a different string than the `http://r3.o.lencr.org` which is the OCSP responder which is already being extracted it seems (as `certinfo.ocsp_uri`).

benaryorg:
The `issuer_uri` is retrieved from the Authority Information Access field the same way as the OCSP responder URI is. Handling is exactly the same since they reside in the same OID space and have the same data type. Tests have also been added based on the integration test certificates.

Signed-off-by: benaryorg <binary@benary.org>

##### SUMMARY

Fixes #529 

Note: this code was semi-hastily put together after I saw that the OCSP responder handling was already there (and it is pretty much the same), so if I missed any additional test cases that should be added or anything else, please say something.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`x509_certificate_info`

##### ADDITIONAL INFORMATION

```yaml
- name: Get certificate information
  community.crypto.x509_certificate_info:
    path: /some/certificate.pem
  register: certinfo

- name: Get upstream certificate
  ansible.builtin.debug:
    # this will print the URI at which the issuer certificate can be retrieved, useful for rebuilding certificate chains.
    msg: |-
      {{ certinfo.issuer_uri }}
```

Before this change no `issuer_uri` will be present, with the change it will be present (for further information consult the added test case).

### DIFF
--- a/plugins/modules/x509_certificate_info.py
+++ b/plugins/modules/x509_certificate_info.py
@@ -378,6 +378,12 @@ ocsp_uri:
                  C(none) if no OCSP responder URI is included.
     returned: success
     type: str
+issuer_uri:
+    description: The Issuer URI, if included in the certificate. Will be
+                 C(none) if no issuer URI is included.
+    returned: success
+    type: str
+    version_added: 2.9.0
 '''
 
 
--- a/changelogs/fragments/aia_issuer.yaml
+++ b/changelogs/fragments/aia_issuer.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - x509_certificate_info - adds ``issuer_uri`` field in return value based on Authority Information Access data (https://github.com/ansible-collections/community.crypto/pull/530).
--- a/plugins/module_utils/crypto/module_backends/certificate_info.py
+++ b/plugins/module_utils/crypto/module_backends/certificate_info.py
@@ -139,6 +139,10 @@ class CertificateInfoRetrieval(object):
     def _get_ocsp_uri(self):
         pass
 
+    @abc.abstractmethod
+    def _get_issuer_uri(self):
+        pass
+
     def get_info(self, prefer_one_fingerprint=False):
         result = dict()
         self.cert = load_certificate(None, content=self.content, backend=self.backend)
@@ -200,6 +204,7 @@ class CertificateInfoRetrieval(object):
         result['serial_number'] = self._get_serial_number()
         result['extensions_by_oid'] = self._get_all_extensions()
         result['ocsp_uri'] = self._get_ocsp_uri()
+        result['issuer_uri'] = self._get_issuer_uri()
 
         return result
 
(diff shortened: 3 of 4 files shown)


felixfontein (commented):
Thank you for your contribution! Can you also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Besides that, it looks good!

felixfontein:
recheck

benaryorg:
> Thank you for your contribution! Can you also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Besides that, it looks good!

I've never written a changlog fragment before, so please let me know if that's correct.
Thank you very much for the quick review too!

felixfontein (approved):

plugins/modules/x509_certificate_info.py
```diff
@@ -378,6 +378,11 @@
                  C(none) if no OCSP responder URI is included.
     returned: success
     type: str
+issuer_uri:
+    description: The Issuer URI, if included in the certificate. Will be
+                 C(none) if no issuer URI is included.
+    returned: success
+    type: str
```
felixfontein, line 385:
```suggestion
    type: str
    version_added: 2.9.0
```

felixfontein (commented):
Sorry, forgot one detail:

felixfontein (approved):

felixfontein:
> Thank you very much for the quick review too!

You're lucky that this landed in my lunchbreak when working from home, otherwise it would have taken longer ;)

felixfontein:
Please ignore the failing sanity test, it should no longer fail on `main` now so I'm going to ignore it.

felixfontein:
@benaryorg thanks again for your contribution!

benaryorg:
> Please ignore the failing sanity test, it should no longer fail on `main` now so I'm going to ignore it.

Oh well, I just had the fix written for that xD
Thanks again for the quick merge!

felixfontein:
@benaryorg this is now contained in today's 2.9.0 release.

PR #541, opened 2022-12-11
CI: extend matrix to run more tests with VMs

felixfontein:
##### SUMMARY
Switches some Docker containers to VMs (Ubuntu 20.04, 22.04; Fedora 36; Alpine 1.16), and adds CentOS Stream 8 with Python 3.6 (system Python interpreter).

CC @mattclay (as I don't know whether these VMs are supposed to be used by collections as well)

CC @Spredzy @Andersson007 as this extends the CI matrix (even though just minimally) and RH is paying for that :)

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -109,16 +109,17 @@ stages:
           targets:
             - name: CentOS 7
               test: centos7
-            - name: Fedora 36
-              test: fedora36
             - name: openSUSE 15
               test: opensuse15
-            - name: Ubuntu 20.04
-              test: ubuntu2004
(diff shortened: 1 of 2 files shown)


felixfontein:
@mattclay the main reason for VMs is that some tests do not work in containers. In this collection, there's only one though, namely the ones for the `luks_device` module: it needs to create loopback devices, which doesn't work (it could work with `--docker-privileged`, but that's something I don't really want to use in CI).

Maybe it's possible to solve this in another way though, but it would probably need support from ansible-test's side as well.

felixfontein:
Closing in favor of #545.

PR #544, opened 2022-12-20
CI: improve CI matrix, split into two groups

felixfontein:
##### SUMMARY
Remove some duplicate targets, and instead add CentOS Stream 8 with Python 3.6.

Also splits CI up into two groups, so that a full run can go through faster. Right now (all tests in a row) can run up to 30 minutes, which is quite long.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
The OpenSuSE problems seem to be unrelated to this PR.

PR #550, opened 2022-12-28
Add filter plugins for (selected) _info modules

felixfontein:
##### SUMMARY
For #542 (see https://github.com/ansible-collections/community.crypto/issues/542#issuecomment-1365935470).

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
filters

PR #554, opened 2022-12-30
Add openssl_csr_info filter

felixfontein:
##### SUMMARY
Part of #550. For #542 (see https://github.com/ansible-collections/community.crypto/issues/542#issuecomment-1365935470).

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
openssl_csr_info

### DIFF
--- a/plugins/modules/openssl_csr_info.py
+++ b/plugins/modules/openssl_csr_info.py
@@ -50,8 +50,12 @@ options:
         choices: [ auto, cryptography ]
 
 seealso:
-- module: community.crypto.openssl_csr
-- module: community.crypto.openssl_csr_pipe
+    - module: community.crypto.openssl_csr
+    - module: community.crypto.openssl_csr_pipe
+    - ref: community.crypto.openssl_csr_info filter <ansible_collections.community.crypto.openssl_csr_info_filter>
+    # - plugin: community.crypto.openssl_csr_info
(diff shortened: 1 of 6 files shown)


plugins/modules/openssl_csr_info.py
```diff
@@ -50,8 +50,12 @@
         choices: [ auto, cryptography ]
 
 seealso:
-- module: community.crypto.openssl_csr
-- module: community.crypto.openssl_csr_pipe
+    - module: community.crypto.openssl_csr
+    - module: community.crypto.openssl_csr_pipe
+    - ref: community.crypto.openssl_csr_info filter <ansible_collections.community.crypto.openssl_csr_info_filter>
+    # - plugin: community.crypto.openssl_csr_info
+    #   plugin_type: filter
```
briantist, line 57: when will these be safe to uncomment?

briantist (approved):

plugins/modules/openssl_csr_info.py
felixfontein: Once we have semantic markup support... (We've sneaked that one into the semantic markup spec, since supporting this requires support from all docs consumers as well, and the semantic markup spec already has `P(...)` for referencing other plugin types.)

felixfontein:
@briantist thanks for reviewing!

PR #555, opened 2022-12-30
Add openssl_privatekey_info filter

felixfontein:
##### SUMMARY
Part of #550. For #542 (see https://github.com/ansible-collections/community.crypto/issues/542#issuecomment-1365935470).

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
openssl_privatekey_info

### DIFF
--- a/plugins/filter/openssl_privatekey_info.py
+++ b/plugins/filter/openssl_privatekey_info.py
@@ -1,0 +1,193 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2022, Felix Fontein <felix@fontein.de>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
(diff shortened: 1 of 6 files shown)


plugins/filter/openssl_privatekey_info.py
```diff
@@ -0,0 +1,193 @@
...
+        required: true
+    passphrase:
+        description:
+            - The passphrase for the private key.
+        type: str
+    return_private_key_data:
+        description:
+            - Whether to return private key data.
+            - Only set this to C(true) when you want private information about this key to
+              leave the remote machine.
```
briantist, line 33: This description could probably use an update since there is no target machine in the case of a filter.

briantist (approved):

plugins/filter/openssl_privatekey_info.py
felixfontein: Thanks for spotting, I missed that one... I've updated it in bd9e1cc24df1339b68e250573e0bb686fc83acf8

briantist (approved):

felixfontein:
@briantist thanks for reviewing this one as well!

PR #569, opened 2023-02-04
Remove unneccessary imports

felixfontein:
##### SUMMARY
Prepare for https://github.com/ansible-collections/news-for-maintainers/issues/34.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
various

### DIFF
--- a/plugins/plugin_utils/action_module.py
+++ b/plugins/plugin_utils/action_module.py
@@ -69,9 +69,9 @@ try:
     # For ansible-core 2.11, we can use the ArgumentSpecValidator. We also import
     # ModuleArgumentSpecValidator since that indicates that the 'classical' approach
     # will no longer work.
-    from ansible.module_utils.common.arg_spec import (
+    from ansible.module_utils.common.arg_spec import (  # pylint: disable=unused-import
         ArgumentSpecValidator,
-        ModuleArgumentSpecValidator,  # noqa
+        ModuleArgumentSpecValidator,
     )
     from ansible.module_utils.errors import UnsupportedError
     HAS_ARGSPEC_VALIDATOR = True
--- a/plugins/plugin_utils/action_module.py
+++ b/plugins/plugin_utils/action_module.py
@@ -71,7 +71,7 @@ try:
     # will no longer work.
     from ansible.module_utils.common.arg_spec import (  # noqa: F401, pylint: disable=unused-import
         ArgumentSpecValidator,
-        ModuleArgumentSpecValidator,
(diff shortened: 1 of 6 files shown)


plugins/plugin_utils/action_module.py
```diff
@@ -69,9 +69,9 @@
     # For ansible-core 2.11, we can use the ArgumentSpecValidator. We also import
     # ModuleArgumentSpecValidator since that indicates that the 'classical' approach
     # will no longer work.
-    from ansible.module_utils.common.arg_spec import (
+    from ansible.module_utils.common.arg_spec import (  # noqa: F401, pylint: disable=unused-import
```
kristianheljas, line 72: `ArgumentSpecValidator` (and any future additions) doesn't seem to require the exception, could splitting this to two separate imports make sense?

```python
from ansible.module_utils.common.arg_spec import ArgumentSpecValidator
from ansible.module_utils.common.arg_spec import ModuleArgumentSpecValidator  # noqa: F401, pylint: disable=unused-import
```

felixfontein: Hmm, I'm totally unsure about this one. I don't really like to have two import statements from the same module. ([ESLint](https://en.wikipedia.org/wiki/ESLint) for example disallows this by default, I think.) On the other hand, I like that ignores ignore only what they should really ignore.

I've been thinking about this since yesterday and I'm still not sure what to prefer :) Maybe someone else has an opinon as well?

kristianheljas: I also don't have strong opinion about this so whatever the maintainer decides is fine by me.
If you decide to keep the single import, adding a comment what actually needs ignoring might be worthwhile.

13th is only getting closer, so I'm gonna approve this as is for now.

kristianheljas (approved):
There doesn't seem to a good solution to [ignoring single name](https://github.com/ansible-collections/community.crypto/pull/569/files#r1098527315) from an import statement, it's fine as it is.

Everything else looks good!

plugins/plugin_utils/action_module.py
felixfontein: I've added a comment in 8543dc3783b5ab26b7ddb059c9d44708f674fb66.

kristianheljas (approved):
Thanks @felixfontein for making it clear what needs to be ignored, LGTM!

plugins/plugin_utils/action_module.py
kristianheljas: /resolved

felixfontein:
@kristianheljas thanks a lot for reviewing this!

PR #572, opened 2023-02-09
action plugin helper: fix deprecation handling

felixfontein:
##### SUMMARY
This is a consequence of https://github.com/ansible/ansible/pull/79681 and https://github.com/ansible/ansible/pull/79740, which were backported to stable-2.14 and released in ansible-core 2.14.2. Without these, there will be a crash when deprecations appear with ansible-core 2.14.2+ in action plugins using this helper.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/plugin_utils/action_module.py

### DIFF
--- a/changelogs/fragments/572-action-module.yml
+++ b/changelogs/fragments/572-action-module.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "action plugin helper - fix handling of deprecations for ansible-core 2.14.2 (https://github.com/ansible-collections/community.crypto/pull/572)."
--- a/plugins/plugin_utils/action_module.py
+++ b/plugins/plugin_utils/action_module.py
@@ -145,9 +145,14 @@ class AnsibleActionModule(object):
             # warnings and deprecations that do not work in plugins. This is a copy of that code adjusted
             # for our use-case:
             for d in self._validation_result._deprecations:
-                self.deprecate(
-                    "Alias '{name}' is deprecated. See the module docs for more information".format(name=d['name']),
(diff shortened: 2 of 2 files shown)


kristianheljas (approved):
I kinda hate that there's a path now where deprecations could in theory be ignored (if `name` and `msg` both get missing), previously any scenario would be caught..

But it is [consistent with ansible-core](https://github.com/ansible/ansible/pull/79681/files#diff-c976a49aa4f2eaea75c52ed5db33f2d622ba55fbc969a8909a8e849be64c62abR295-R302), so .. LGTM!

felixfontein:
> But it is [consistent with ansible-core](https://github.com/ansible/ansible/pull/79681/files#diff-c976a49aa4f2eaea75c52ed5db33f2d622ba55fbc969a8909a8e849be64c62abR295-R302), so .. LGTM!

It's not anymore, https://github.com/ansible/ansible/pull/79740 changed that code so that `msg` is always there.

> I kinda hate that there's a path now where deprecations could in theory be ignored [...] previously any scenario would be caught..

Previously any scenario with ansible-core 2.14.2+ would result in a crash, I wouldn't exactly call that 'caught' ;-) But yeah, I'm not 100% happy with this. I think (hope) that ansible-core will stick to `msg` from now on, though, so I *think* it should be fine. (I guess, until it is not... but at least then it won't crash)

felixfontein:
@kristianheljas thanks a lot for reviewing this!

Issue #591, opened 2023-04-05
Failure using method (v2_runner_on_ok) in callback plugin ... 'utf-8' codec can't encode character '\udc80' in position 1886: surrogates not allowed


laidbackware:
##### SUMMARY
When attempting to use the example from community.crypto.get_certificate contact google.com, it triggers the exception below.

```
[WARNING]: Failure using method (v2_runner_on_ok) in callback plugin (<ansible.plugins.callback.default.CallbackModule object at 0x7f37ab1034f0>): 'utf-8' codec can't encode character '\udc80' in position 1886: surrogates not allowed
Callback Exception: 
  File "/home/matt/.local/lib/python3.10/site-packages/ansible/executor/task_queue_manager.py", line 450, in send_callback
    method(*new_args, **kwargs)
   File "/home/matt/.local/lib/python3.10/site-packages/ansible/plugins/callback/default.py", line 106, in v2_runner_on_ok
    self._display.display(msg, color=color)
   File "/home/matt/.local/lib/python3.10/site-packages/ansible/utils/display.py", line 289, in display
    fileobj.write(msg2)
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.crypto.get_certificate

##### ANSIBLE VERSION
```
ansible [core 2.14.4]
  config file = None
  configured module search path = ['/home/matt/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/matt/.local/lib/python3.10/site-packages/ansible
  ansible collection location = /home/matt/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/matt/.local/bin/ansible
  python version = 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION
```
# /home/matt/.ansible/collections/ansible_collections
Collection       Version
---------------- -------
community.crypto 2.11.1 

# /home/matt/.local/lib/python3.10/site-packages/ansible_collections
Collection       Version
---------------- -------
community.crypto 2.11.1
```

##### CONFIGURATION
```
CONFIG_FILE() = None
```

##### OS / ENVIRONMENT
Running locally on Ubuntu 22.04 Desktop
Kernel Linux derek 5.19.0-38-generic

##### STEPS TO REPRODUCE
Running the playbook below with `-vvv`, generates the error.
```yaml
[...]

PR #592, opened 2023-04-06
get_certificate: add asn1_base64 option

felixfontein:
##### SUMMARY
This allows to work around the crash in the default callback; see #591.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
get_certificate


briantist (approved):
Looks great to me!

> that value will eventually be deprecated and changed to C(true)

Only question I have is why not set a deprecation date/warning now?

felixfontein:
> Only question I have is why not set a deprecation date/warning now?

So that there are at least a few versions which already have the new option, but do not yet emit the deprecation warning. That allows users to adopt the change in a nicer fashion if they use playbooks/roles not always with the same version of community.crypto.

felixfontein:

PR #596, opened 2023-04-28
x509_crl: prepare releasing the mode option for AnsibleModule's use

felixfontein:
##### SUMMARY
Since https://github.com/ansible/ansible/pull/80449 got merged the module no longer works with ansible-core devel, since the `mode` values `generate` and `update` are not allowed by ansible-core's validation.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
x509_crl

### DIFF
--- a/plugins/modules/x509_crl.py
+++ b/plugins/modules/x509_crl.py
@@ -42,7 +42,7 @@ options:
         default: present
         choices: [ absent, present ]
 
-    mode:
+    crl_mode:
         description:
             - Defines how to process entries of existing CRLs.
             - If set to C(generate), makes sure that the CRL has the exact set of revoked certificates
@@ -51,8 +51,17 @@ options:
               I(revoked_certificates), but can also contain other revoked certificates. If the CRL file
               already exists, all entries from the existing CRL will also be included in the new CRL.
               When using C(update), you might be interested in setting I(ignore_timestamps) to C(true).
+            - The default value is C(generate).
         type: str
-        default: generate
+        # default: generate
+        choices: [ generate, update ]
+        version_added: 2.13.0
+    mode:
+        description:
+            - This parameter is deprecated and will be removed in community.crypto 3.0.0. Use I(crl_mode) instead.
+            - Note that from community.crypto 3.0.0 on, I(mode) will be used for the CRL file's mode.
+        type: str
+        # default: generate
         choices: [ generate, update ]
 
     force:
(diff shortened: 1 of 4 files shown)


felixfontein:
This fixes the nightly CI failures: https://dev.azure.com/ansible/community.crypto/_build/results?buildId=75691&view=results

felixfontein:
Restart CI.

plugins/modules/x509_crl.py
```diff
@@ -51,8 +51,17 @@
...
               already exists, all entries from the existing CRL will also be included in the new CRL.
               When using C(update), you might be interested in setting I(ignore_timestamps) to C(true).
+            - The default value is C(generate).
+        type: str
+        # default: generate
+        choices: [ generate, update ]
+        version_added: 2.13.0
+    mode:
+        description:
+            - This parameter is deprecated and will be removed in community.crypto 3.0.0. Use I(crl_mode) instead.
```
briantist, line 61: It looks like this option is the new `crl_mode` option, as defined in line 45, but it's being set at deprecated with a message to use `crl_mode` "instead". Am I misreading something?

felixfontein: Yes; basically I renamed `mode` to `crl_mode`, but instead of just adding `mode` as an alias to `crl_mode` I have to keep it as a separate option, since otherwise `mode` as an option would show up from `AnsibleModule`'s `add_file_common_args=True`.

felixfontein: How about changing the documentation to:
```suggestion
            - This parameter has been renamed to I(crl_mode). The old name I(mode) is now deprecated and will be removed in community.crypto 3.0.0.
              Simply replace usage of this parameter with I(crl_mode).
```

plugins/modules/x509_crl.py
```diff
@@ -51,8 +51,17 @@
               I(revoked_certificates), but can also contain other revoked certificates. If the CRL file
               already exists, all entries from the existing CRL will also be included in the new CRL.
               When using C(update), you might be interested in setting I(ignore_timestamps) to C(true).
+            - The default value is C(generate).
```
felixfontein, line 54: And another possible update here:
```suggestion
            - The default value is C(generate).
            - This parameter was called I(mode) before community.crypto 2.13.0. It has been renamed to avoid
              a collision with the common I(mode) parameter for setting the CRL file's access mode.
```

plugins/modules/x509_crl.py
briantist: I see my mistake, I completely missed line 59. What had me confused was thinking that 61 was still part of `crl_mode` 🤦 

With that, the old description was fine. The new description is good too; only thing I'd avoid is use of the word "simply" but maybe that's my own feeling only.

felixfontein: I removed 'simply', see 6666bafdcd2c90610c1e997d7432405b1248d036.

briantist (approved):

felixfontein:

PR #606, opened 2023-05-19
Rewrite EE test workflows to use ansible-builder 3.0.0; fix EE dependencies

felixfontein:
##### SUMMARY
Also use other base images. ansible-builder 3.0.0 allows arbitrary "RHEL-like" base images, as opposed to ansible-builder < 3.0.0 (which required very specially prepared base images).

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
EE tests


felixfontein:
CC @Andersson007

felixfontein:
It seems we're hitting https://github.com/ansible/ansible-navigator/issues/1448 when using podman. Let's see whether switching to docker helps... (it did for me locally.)

nitzmahone:
Hrm, with a reasonably recent runner version and the default builder 3.0 config, it should have the `awx_display` bits copied in at runtime and configured dynamically by runner, and the entrypoint script should ensure that the container user always has a writeable homedir (so long as it's running as GID0 inside). If navigator's trying too hard to force both the host user and gid, it's probably not going to end well, since we have no way for the entrypoint to provide a writeable homedir in that case.

felixfontein:
ready_for_review

felixfontein:
https://github.com/ansible/ansible-navigator/pull/1539 fixed the podman issue, I've successfully switched back from docker to podman now that that fix is released (ansible-navigator 3.3.1).

Andersson007:
@felixfontein thanks for pinging! Please do it in related PRs (or maybe even better to put them also to the EE feedback issue to have them in one place)
once I finish with the basic overviews / how-tos, i'll take a look;)

PR #612, opened 2023-05-31
[WIP] Ansible-lint: Add example config for github action

### LINKED ISSUE #611: Add Ansible-lint GitHub action
##### SUMMARY
Ansible-lint want to have their linter (and probably also their github action) used throughout the ecosystem a bit more. We could/should run it on PRs and the main/release branches.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
Tests written in Ansible-YAML ( I don't think they currently are very concerned with Python code, that's more the area of ansible-test)

##### ADDITIONAL INFORMATION
See https://github.com/ansible/ansible-lint/issues/3482

MarkusTeufelberger:
##### SUMMARY
resolves #611 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
Tests

##### ADDITIONAL INFORMATION
Let's see how this works out.


MarkusTeufelberger:
There's also https://github.com/ansible/ansible-lint-action, maybe that's what they meant to use instead? Let's try...

MarkusTeufelberger:
Ah well... "CRITICAL Linter finished without analyzing any file, check configuration and arguments given."

Seems like it still has its issues when running against a collection even in a very default layout.

felixfontein:
The linter mainly lints the YAML integration tests in tests/integration/targets/. It will produce a HUGE number of errors/warnings/...

My suggestion would be to simply ignore all errors for now.

(But for that it first has to actually run...)

.github/workflows/ansible-lint.yml
```diff
@@ -0,0 +1,16 @@
...
+on:
+  pull_request:
+    branches: ["main", "stable-*"]
+jobs:
+  build:
+    name: Ansible Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run ansible-lint
+        uses: ansible/ansible-lint-action@v6
```
felixfontein, line 16: You probably need to check out the repository first, so the linter has something to run on :)

MarkusTeufelberger: I would have guessed so too, the official documentation says otherwise: https://ansible-lint.readthedocs.io/installing/#installing-ansible-lint-as-a-github-action

MarkusTeufelberger:
Wow, it is almost impressive how much it complains!

.github/workflows/ansible-lint.yml
felixfontein: The official documentation seems to be pretty out of date.

MarkusTeufelberger:
I guess I can add ignores for everything and then start fixing stuff one-by-one either automatically (via `ansible-lint --write`) or by hand if automatic rewrites are not easy or available. At least the long tail of single digit errors should be doable in this PR I hope.

.ansible-lint
```diff
@@ -0,0 +1,31 @@
+---
+skip_list:
+  # TODO: Fix and enable one-by-one over time
```
felixfontein, line 3: But only the ones that make sense. `sanity[cannot-ignore]` looks pretty much wrong, for example.

.ansible-lint
```diff
@@ -0,0 +1,31 @@
+---
```
felixfontein, line 1:
```suggestion
---
# Copyright (c) Ansible Project
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later

```

.ansible-lint
MarkusTeufelberger: I hope that the eventual goal is to have both the collection(s) and ansible-lint rules in a state that no rule skips or any changes to default configs are necessary. For `sanity[cannot-ignore]` for example I am hardlocked on https://github.com/ansible/ansible/pull/79700 making it into ansible-core _and_ every version of ansible-core falling out of support until I can disable this in my collection. Similar with the shebangs in this repo here apparently?

There are some easy wins with fixing the long tail of errors (end of line whitespace and such) and automatic rewrites could be used or introduced for some things too. There are however also some issues with that (e.g. https://github.com/ansible/ansible-lint/issues/3226 will hit every collection created from the default template).

MarkusTeufelberger:
I ran against 6.17.0 locally and it complains even more. Added a few quick fixes (`on:`in github workflow files is a truthy name unfortunately and `order` apparently is a reserved name by Ansible these days) and 2 more ignores for upcoming rules that were also quite noisy. Now I can also locally start disabling ignores one by one and fix them. First though I also want to get ansible-lint to be idempotent in `--write` mode (fixing or "fixing" lots of YAML formatting usually). That alone looks like it'll be a horrible thing to review and likely we'll be stuck with an `.ansible-lint` file in the root directory because it tries to "fix" the github workflow files too.

.github/workflows/ansible-test.yml
```diff
@@ -7,7 +7,7 @@
 # https://github.com/marketplace/actions/ansible-test
 
 name: EOL CI
-on:
+on:  # yamllint disable-line rule:truthy
```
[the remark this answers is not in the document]
MarkusTeufelberger: Hm, `ansible-lint --write` seems to prefer `"on"` but it works, yes. :-)

MarkusTeufelberger:
Looks like they want to use the action from https://github.com/ansible/ansible-lint/blob/main/action.yml actually, not the one from that other repository. I'll update this PR next week thusly.

The other common change in other collections seems to be to use an `.ansible-lint-ignore` file (https://ansible.readthedocs.io/projects/lint/configuring/#ignoring-rules-for-entire-files) instead of a config file in the root of the repository so new contributions would always have to adhere to all rules instead of allowing "bad" code in and fixing all instances of a rule violation eventually. There doesn't seem to be `--write` mode applied though.

felixfontein:
Fine with me, as long as `ignore-errors` and `sanity[cannot-ignore]` stay permanently disabled in `.ansible-lint` .

MarkusTeufelberger:
I'm not sure if it is better to kinda document all 533 rule violations and eventually fix them while not allowing new ones because there's not much to take inspiration from (aka. copy paste) when writing new tests for example. I don't see how current violations are much better or worthy of keeping around though on the other hand it just makes more work for the future to eventually fix all this rule by rule. In any case it is easy enough (and just tedious) to fix stuff, it would just be nice if we could actually use `--write` mode for this which is currently completely out of question. I now pushed the "minimal rule ignore config + centrally documented ignores + lint config as given in the main repo" version as a comparison.

felixfontein:
Also considering that .ansible-lint-ignore is 41 KiB, I think the previous version of the PR was nicer - mainly for the reasons you mentioned :-)

.github/workflows/ansible-lint.yml
```diff
@@ -0,0 +1,16 @@
...
+name: ansible-lint
+"on":
+  pull_request:
+    branches: ["main", "stable-*"]
+jobs:
+  build:
+    name: Ansible Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run ansible-lint
```
felixfontein, line 15: A checkout step needs to be added before here. There was a breaking change in the action that started requiring that.

felixfontein:
@MarkusTeufelberger do you want to further update this? Or should we restart this with a fresh PR? (Or abandon it for now?)

MarkusTeufelberger:
I think I'll rebase and check with the current version.

felixfontein:
I guess #886 should have closed this one as well.

PR #622, opened 2023-06-10
Added support for certificates in DER format for `x509_certificate_info` module

### LINKED ISSUE #603: DER format - `community.crypto.x509_certificate_info`
##### SUMMARY

Currently only PEM format is supported for `community.crypto.x509_certificate_info` module, so when we try get info about certificate in DER format then we getting:
```
"msg": "Unable to load PEM file. See https://cryptography.io/en/latest/faq/#why-can-t-i-import-my-pem-file for more details. MalformedFraming"
```

I noticed `community.crypto.x509_certificate_info` module uses `cryptography`. Current code uses `load_pem_x509_certificate` method but there is a `load_der_x509_certificate` method in `cryptography` to load DER certificates.
It would be better than using `openssl x509 -in cert.crt -inform der -text -noout` command to get certificate info.
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

community.crypto.x509_certificate_info
##### ADDITIONAL INFORMATION

https://github.com/ansible-collections/community.crypto/blob/e4ebca09452d78046618f67208405652ce6cc889/plugins/module_utils/crypto/support.py#L207
https://cryptography.io/en/latest/x509/reference/#cryptography.x509.load_der_x509_certificate

```yaml

```

msl0:
##### SUMMARY

Resolves #603 by adding support for DER-format certificates when loading a certificate via `path`.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`community.crypto.x509_certificate_info`

##### ADDITIONAL INFORMATION

I have not found a solution to load the certificate via the `content` parameter. It fails due to module input encoding (`ansible_community.crypto.x509_certificate_info_payload.zip/ansible/module_utils/basic.py\", line 1392, in _log_invocation\nUnicodeEncodeError: 'utf-8' codec can't encode character '\\udc82' in position 1: surrogates not allowed`), but in my opinion it's still better than calling openssl command externally.

```paste below

```

### DIFF
--- a/plugins/module_utils/crypto/support.py
+++ b/plugins/module_utils/crypto/support.py
@@ -203,7 +203,7 @@ def load_certificate(path, content=None, backend='cryptography'):
     if backend == 'pyopenssl':
         return crypto.load_certificate(crypto.FILETYPE_PEM, cert_content)
     elif backend == 'cryptography':
-        if any([x in cert_content for x in [b'-----BEGIN CERTIFICATE-----', b'-----END CERTIFICATE-----\n']]):
+        if any([x in cert_content for x in [b'-----BEGIN CERTIFICATE-----', b'-----END CERTIFICATE-----']]):
             try:
                 return x509.load_pem_x509_certificate(cert_content, cryptography_backend())
             except ValueError as exc:
--- a/plugins/module_utils/crypto/support.py
+++ b/plugins/module_utils/crypto/support.py
@@ -203,10 +203,16 @@ def load_certificate(path, content=None, backend='cryptography'):
     if backend == 'pyopenssl':
         return crypto.load_certificate(crypto.FILETYPE_PEM, cert_content)
     elif backend == 'cryptography':
-        try:
-            return x509.load_pem_x509_certificate(cert_content, cryptography_backend())
-        except ValueError as exc:
-            raise OpenSSLObjectError(exc)
+        if any([x in cert_content for x in [b'-----BEGIN CERTIFICATE-----', b'-----END CERTIFICATE-----\n']]):
+            try:
+                return x509.load_pem_x509_certificate(cert_content, cryptography_backend())
+            except ValueError as exc:
+                raise OpenSSLObjectError(exc)
+        else:
+            try:
+                return x509.load_der_x509_certificate(cert_content, cryptography_backend())
+            except ValueError as exc:
+                raise OpenSSLObjectError('Cannot parse DER certificate: {0}'.format(exc))
 
 
 def load_certificate_request(path, content=None, backend='cryptography'):
--- a/plugins/modules/x509_certificate_info.py
+++ b/plugins/modules/x509_certificate_info.py
@@ -40,10 +40,11 @@ options:
         description:
             - Remote absolute path where the certificate file is loaded from.
             - Either I(path) or I(content) must be specified, but not both.
+            - PEM and DER formats are supported.
         type: path
     content:
         description:
-            - Content of the X.509 certificate in PEM or DER format.
+            - Content of the X.509 certificate in PEM format.
             - Either I(path) or I(content) must be specified, but not both.
         type: str
         version_added: '1.0.0'
--- a/plugins/modules/x509_certificate_info.py
+++ b/plugins/modules/x509_certificate_info.py
@@ -43,7 +43,7 @@ options:
         type: path
     content:
         description:
-            - Content of the X.509 certificate in PEM format.
+            - Content of the X.509 certificate in PEM or DER format.
             - Either I(path) or I(content) must be specified, but not both.
         type: str
         version_added: '1.0.0'


plugins/module_utils/crypto/support.py
```diff
@@ -203,10 +203,16 @@ def load_certificate(path, content=None, backend='cryptography'):
     if backend == 'pyopenssl':
         return crypto.load_certificate(crypto.FILETYPE_PEM, cert_content)
     elif backend == 'cryptography':
-        try:
-            return x509.load_pem_x509_certificate(cert_content, cryptography_backend())
-        except ValueError as exc:
-            raise OpenSSLObjectError(exc)
+        if any([x in cert_content for x in [b'-----BEGIN CERTIFICATE-----', b'-----END CERTIFICATE-----']]):
```
felixfontein, line 206: Please use `identify_pem_format()` from `ansible_collections.community.crypto.plugins.module_utils.crypto.pem`.

Also please add a new parameter to this function which allows to enable this feature (default must be to not enable it), and only use this in `x509_certificate_info`. Otherwise you are potentially breaking other code which relies on this function to not load DER certificates. (Also if `backend == 'pyopenssl'` when this feature is enabled, fail.)

felixfontein (commented):
Thanks for your contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

plugins/module_utils/crypto/support.py
msl0: Thank you for your suggestion. I added the suggested changes.

plugins/module_utils/crypto/support.py
```diff
@@ -201,12 +205,21 @@ def load_certificate(path, content=None, backend='cryptography'):
     except (IOError, OSError) as exc:
         raise OpenSSLObjectError(exc)
     if backend == 'pyopenssl':
-        return crypto.load_certificate(crypto.FILETYPE_PEM, cert_content)
+        if identify_pem_format(cert_content) or der_support_enabled is False:
```
felixfontein, line 208: I would write it this way around:
```suggestion
        if der_support_enabled is False or identify_pem_format(cert_content):
```
then `identify_pem_format(cert_content)` is only ever called if `der_support_enabled is not False`.

plugins/module_utils/crypto/support.py
```diff
@@ -201,12 +205,21 @@ def load_certificate(path, content=None, backend='cryptography'):
...
+            return crypto.load_certificate(crypto.FILETYPE_PEM, cert_content)
+        elif der_support_enabled:
+            raise OpenSSLObjectError('Certificate in DER format is not supported by the pyopenssl backend.')
     elif backend == 'cryptography':
-        try:
-            return x509.load_pem_x509_certificate(cert_content, cryptography_backend())
-        except ValueError as exc:
-            raise OpenSSLObjectError(exc)
+        if identify_pem_format(cert_content) or der_support_enabled is False:
```
felixfontein, line 213:
```suggestion
        if der_support_enabled is False or identify_pem_format(cert_content):
```

msl0: The order was intentional to preserve the original behavior (raise PEM-format exception) when specifying a certificate in DER format indicating der_support_enabled to true. When we set der_support_enabled to True is means we can provide certificate in PEM or DER format - both are supported. If we don't care about preserving the original behavior, then we could apply your suggestion but we need to change `or` to `and` to support both.
I hope that clarified why I did it. What do you recommend?

plugins/modules/x509_certificate_info.py
```diff
@@ -447,7 +448,7 @@ def main():
             valid_at[k] = get_relative_time_option(v, 'valid_at.{0}'.format(k))
 
     try:
-        result = module_backend.get_info()
+        result = module_backend.get_info(der_support_enabled=True)
```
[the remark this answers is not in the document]
msl0: I don't think it changes anything when trying to load a DER certificate via `content`. In my opinion the code stops at an earlier stage while loading the parameters (problem with encoding to utf-8).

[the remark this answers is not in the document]
felixfontein: You are assuming that a valid DER is passed. The user can also pass in random garbage that happens to be valid UTF-8, in which case it tries to load it as a DER and provides a different error message than before.

plugins/module_utils/crypto/support.py
felixfontein: Right now even if the caller provides `der_support_enabled == False`, `identify_pem_format()` will be executed, even though its return value will not be used. Please change the order so that `identify_pem_format()` will not be executed when `der_support_enabled == False`.

Your behavior introduces additional overhead then `der_support_enabled == False`, while mine restores the original behavior (which is do not run `identify_pem_format()`).

felixfontein:
@msl0 thanks a lot for your contribution!

PR #640, opened 2023-07-15
Fix license disclaimer for some vendored Jinja2 code in tests

felixfontein:
##### SUMMARY
The code (mostly) comes from Jinja2, so the correct license declaration needs to be used.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
tests/integration/targets/prepare_jinja2_compat/filter_plugins/jinja_compatibility.py


felixfontein:
CC @gotmax23

felixfontein:
recheck

tests/integration/targets/prepare_jinja2_compat/filter_plugins/jinja_compatibility.py
```diff
@@ -1,6 +1,11 @@
-# Copyright (c) Ansible Project
-# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
-# SPDX-License-Identifier: GPL-3.0-or-later
+# This code is part of Ansible, but is an independent component.
+# This particular file snippet, and this file snippet only, is licensed under the
+# BSD-3-Clause  License. Modules you write using this snippet, which is embedded
+# dynamically by Ansible, still belong to the author of the module, and may assign
+# their own license to the complete work.
```
gotmax23, line 5: I'd prefer if we got rid of this message. I'm unsure about

```
Modules you write using this snippet, which is embedded
# dynamically by Ansible, still belong to the author of the module, and may assign
# their own license to the complete work.
```

and I think the whole disclaimer is unnecessary in general. We can just use the existing three line template.

In any case, we do need to remove the GPL-3.0-or-later header.

tests/integration/targets/prepare_jinja2_compat/filter_plugins/jinja_compatibility.py
```diff
@@ -1,6 +1,11 @@
-# Copyright (c) Ansible Project
-# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
-# SPDX-License-Identifier: GPL-3.0-or-later
+# This code is part of Ansible, but is an independent component.
```
gotmax23, line 1: Also, `part of Ansible` is not entirely actuate.

tests/integration/targets/prepare_jinja2_compat/filter_plugins/jinja_compatibility.py
felixfontein: This is the standard template for vendored code, used all over various collections and ansible-core. (Not in all places, as you noticed, but in many.) We can change it everywhere, but we should do that in a coordinated effort after finding out how exactly this should be written.

tests/integration/targets/prepare_jinja2_compat/filter_plugins/jinja_compatibility.py
felixfontein: This is what we use in many community collections. I think this PR is the wrong time and place to change that formulation.

gotmax23 (approved):
Ack. We can discuss the license notice format in a follow up.

PR #739, opened 2024-04-30
acme_certificate: allow to request renewal of a certificate according to ARI

felixfontein:
##### SUMMARY
See https://www.ietf.org/archive/id/draft-ietf-acme-ari-03.html#name-extensions-to-the-order-obj and https://letsencrypt.org/2024/04/25/guide-to-integrating-ari-into-existing-acme-clients.html#step-6-indicating-which-certificate-is-replaced-by-this-new-order.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
acme_certificate

### DIFF
--- a/plugins/modules/acme_certificate.py
+++ b/plugins/modules/acme_certificate.py
@@ -297,9 +297,13 @@ options:
     description:
       - Determines whether to request renewal of an existing certificate according to
         L(the ACME ARI draft 3, https://www.ietf.org/archive/id/draft-ietf-acme-ari-03.html#section-5).
-      - This is only used if the certificate specified in O(dest) or O(fullchain_dest) already exists.
+      - This is only used when the certificate specified in O(dest) or O(fullchain_dest) already exists.
       - V(never) never sends the certificate ID of the certificate to renew. V(always) will always send it.
       - V(when_ari_supported) only sends the certificate ID if the ARI endpoint is found in the ACME directory.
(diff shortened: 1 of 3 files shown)


plugins/modules/acme_certificate.py
```diff
@@ -293,6 +293,30 @@
...
+      - Generally you should use V(when_ari_supported) if you know that the ACME service supports a compatible
+        draft (or final version, once it is out) of the ARI extension. V(always) should never be necessary.
+        If you are not sure, or if you receive strange errors on invalid C(replaces) values in order objects,
+        use V(never), which also happens to be the default.
+      - ACME servers might refuse to create new orders with C(replaces) for certificates that already have an
+        existing order. This can happen if this module is used to create an order, and then the playbook/role
+        fails in case the challenges cannot be set up. If the playbook/role does not record the order data to
+        continue with the existing order, but tries to create a new one on the next run, creating the new order
+        might fail. For this reason, this option should only be set to a value different from V(never) if the
+        role/playbook using it keeps track of order data accross restarts.
```
felixfontein, line 312: I'm currently thinking of how to better handle this situation. The best way would to make it possible to disable orders, which cannot be done directly, but by disabling its authorizations. (If you don't want to wait for the order to expire.) I'll provide a solution for that in another PR, and in that PR I will also update this description accordingly.

(It might be a good idea to split this module into multiple modules anyway, especially the two-step process makes this module really quirky... But that's another topic :) )

PR #743, opened 2024-05-01
 New gpg_keypair module to manage GPG keys.

### LINKED ISSUE #238: GPG module and key lookup
_From @ageis on May 09, 2019 11:17_

##### SUMMARY
Ansible could use a GPG module, for managing keyrings, importing and exporting keys, verifying and making signatures, and decrypting and encrypting data (both symmetric and non). Working with GPG keys, outside of the context of apt, is an extremely common sysadmin task which we can make it easier to automate. I think the addition of this module could be complemented by a "gpgkey" lookup which returns matching keys. I am on the fence about whether encryption, decryption, and sign+verify functions should be split into their own submodules.

I've seen no prior tickets or discussion of this, so I've taken on the task of developing it myself, and have already implemented a significant amount of the prospective functionality, and am creating this issue in order to track my progress. Please assign it to me. Hopefully this will be ready in time for 2.9. 

This is a project which could prove thorny/hairy to maintain going forward. For elucidation on why that is, I'd love to [quote](https://blog.patternsinthevoid.net/pretty-bad-protocolpeople.html) @isislovecruft: 

> "...largely due to GnuPG’s interface design which uses file descriptors, and only file descriptors, to speak a custom, potentially binary but often ascii, order dependent line protocol, whose line order, keywords, number of fields, and other details are subject to change between minor point versions of GnuPG. If that sounds like a special hell invented by some sort of unholy crossing between RMS and a rabid howler monkey: welcome to working with (or rather, more likely, around) the Terrible Idea Generator known as the GnuPG development team."

For now I've decided to rely on her version of python-gnupg which is now called pretty-bad-protoco] ([PyPi](https://pypi.org/project/pretty-bad-protocol/), [GitHub](https://github.com/isislovecruft/python-gnupg)); in my experience and the opinion of people I trust it's simply better, but that is a decision which the project/community will have to review when it's PR-time. 

I found two modules out there, but IMO neither of them are presently suitable for inclusion in core for various reasons. I want to note the prior effort by @tnt and @brandonkal anyway:
* https://github.com/tnt/ansible-gpg-import-module 
* https://github.com/brandonkal/ansible-gpg

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
- module: gpg
- lookup: gpgkey

##### ADDITIONAL INFORMATION

My work-in-progress is located here: https://git.cointel.pro/FOSS/ansible-module-gpg

_Copied from original issue: ansible/ansible#56262_

austinlucaslake:
##### SUMMARY

New gpg_keypair module to manage GPG keys.

Fixes #238

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

gpg_keypair

##### ADDITIONAL INFORMATION

[Unattended GPG key generation](https://www.gnupg.org/documentation/manuals/gnupg/Unattended-GPG-key-generation.html)

Example usage:

```paste below
community.crypto.gpg_keypair:
  state: present
  key_type: RSA
  key_length: 4096
  subkey_type: RSA
  subkey_length: 4096
  name: {{ name }}
  comment: {{ comment }}
  email: {{ email }}
  passphrase: {{ passphrase }}
```

### DIFF
--- a/plugins/modules/gpg_keypair.py
+++ b/plugins/modules/gpg_keypair.py
@@ -1,0 +1,390 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2024, Austin Lucas Lake <53884490+austinlucaslake@users.noreply.github.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: gpg_keypair
+author: "Austin Lucas Lake (@austinlucaslake)"
+short_description: Generate or delete GPG private and public keys
+description:
+    - "This module allows one to generate or delete OpenSSH private and public keys using GnuPG (gpg)."
+requirements:
+    - gpg >= 2.1
+extends_documentation_fragment:
+    - ansible.builtin.files
+    - community.crypto.attributes
+    - community.crypto.attributes.files
+attributes:
+    check_mode:
+        support: full
+options:
+    state:
+        description:
+            - Whether the private and public keys should exist or not, taking action if the state is different from what is stated.
+        type: str
+        default: present
+        choices: [ present, absent ]
+    key_type:
+        description:
+            - "Specifies the type of key to create. By default this is V(EDDSSA) which must be used with Curve25519.
+              Supported key types are V(RSA), V(DSA), V(ECDSA), V(EDDSA), and V(ECDH)."
+        type: str
+        default: EDDSA
+        choices: ['RSA', 'DSA', 'ECDSA', 'EDDSA', 'ECDH']        
+    key_length:
+        description:
+            - For non-ECC keys, this specifies the number of bits in the key to create.
+            - For RSA keys, the minimum is V(1024), the maximum is V(4096), and the default is V(3072).
+            - For DSA keys, the minimum is V(768), the maximum is V(3072), and the default is V(2048).
+            - Invalid values will automatically be saturated in the afforemented ranges for each respective key.
+            - For ECC keys, this parameter will be ignored.
+        type: int
+    key_curve:
+        description:
+            - For ECC keys, this specifies the curve used to generate the keys.
+            - Supported key curves are V(cv25519), V(nistp256), V(nistp384), V(nistp521), V(brainpoolP256r1), V(brainpoolP384r1), V(brainpoolP512r1), and V(secp256k1).
+            - EDDSA keys can only be used with V(cv25519).
+            - Only EDDSA and ECDH keys support V(cv25519), and for both, V(cv25519) is the default.
+            - For ECDSA and ECDH, the default is V(brainpoolP512r1).
+            - For non-ECC keys, this parameter with be ignored.
+        type: str
+        choices: ['cv25519', 'nistp256', 'nistp384', 'nistp521', 'brainpoolP256r1', 'brainpoolP384r1', 'brainpoolP512r1', 'secp256k1']
+    key_usage:
+        description:
+            - Specifies usage(s) for key.
+            - Support usages are V(encrypt), V(sign), V(auth), V(cert).
+            - V(cert) is given to all primary keys regardess, however can be used to only give V(vert) usage to a key.
+            - If not usage is specified, the valid usages for the given key type with be assigned.
+            - If O(state) is V(absent), this parameter is ignored. 
+        type: list[str]
+        choices: ['encrypt', 'sign', 'auth', 'cert']
+    subkey_type:
+        description:
+            - Similar to O(key_type), but also supports V(ELG).
+        type: str
+        default: EDDSA
+        choices: ['RSA', 'DSA', 'ECDSA', 'EDDSA', 'ECDH', 'ELG']
+    subkey_length:
+        description:
+            - Similar to O(key_length).
+            - For ELG keys, the minimum is V(1024), the maximum is V(4096), and the default is V(3072)."
+        type: int
+    subkey_curve:
+        description:
+            - "Similar to O(key_curve)"
+        type: str
+        choices: ['cv25519', 'nistp256', 'nistp384', 'nistp521', 'brainpoolP256r1', 'brainpoolP384r1', 'brainpoolP512r1', 'secp256k1']
+    key_usage:
+        description:
+            - Similar to O(key_usage), but does not support V(cert).
+        type: list[str]
+        choices: ['encrypt', 'sign', 'auth', 'cert']
+    name:
+        description:
+            - Specifies a name for the key.
+        type: str
+    comment:
+        description:
+            - Specifies a comment for the key.
+        type: str
+    email:
+        description:
+            - Specifies an email for the key.
+        type: str
+    passphrase:
+        description:
+            - Passphrase used to decrypt an existing private key or encrypt a newly generated private key.
+            - If O(state) is V(absent), this parameter is ignored. 
+        type: str
+    fingerprints:
+        description:
+            - Specifies keys to match against.
+            - Provided fingerprints will take priority over user-id "V(name) (V(comment)) <V(email)>".
+            - If O(state) is V(absent), keys with the provided fingerprints will be deleted if found 
+        type: list[str]
+    keyserver:
+        description:
+            - Specifies keyserver to upload key to.
+            - If O(state) is V(absent), this parameter will be ignored.
+        type: str
+    transient_key:
+        description:
+            - Allows key generation to use a faster, but less secure random number generator.
+        type: bool
+        default: False
+    return_fingerprints:
+        description:
+            - Allows for the return of fingerprint(s) for newly created or deleted keys(s)
+        type: bool
+        default: False
+'''
+
+EXAMPLES = '''
+- name: Generate the default GPG keypair (Ed25519)
+  community.crypto.gpg_keypair:
+
+- name: Generate the default GPG keypair with a passphrase
+  community.crypto.gpg_keypair:
+    passphrase: super_secret_password
+
+- name: Generate a RSA GPG keypair with the default RSA size (2048 bits)
+  community.crypto.gpg_keypair:
+    key_type: RSA
+
+- name: Generate a RSA GPG keypair with custom size (4096 bits)
+  community.crypto.gpg_keypair:
+    key_type: RSA
+    key_length: 4096
+
+- name: Generate an ECC GPG keypair 
+  community.crypto.gpg_keypair:
+    key_type: EDDSA
+    key_curve: cv25519
+
+- name: Generate a GPG keypair and with a subkey:
+  community.crypto.gpg_keypair:
+    subkey_type: ECDH
+    subkey_curve: cv25519
+
+- name: Generate a GPG keypair with custom user-id:
+  community.crypto.gpg_keypair:
+    name: Your Name
+    comment: Interesting comment.
+    email: example@email.com
+
+- name: Generate a GPG keypair and return fingerprint of new key
+  community.crypto.gpg_keypair:
+    return_fingerprints: true
+  register: gpg_keys
+
+- name: Delete GPG keypair(s) matching a specified user-id:
+  community.crypto.gpg_keypair:
+    state: abscent
+    name: Your Name
+    comment: Interesting comment.
+    email: example@email.com
+
+- name: Delete GPG keypair(s) matching a specified fingerprint:
+  community.crypto.gpg_keypair:
+    state: abscent
+    fingerprints:
+      - ABC123...
+
+'''
+
+RETURN = '''
+size:
+    description: Size (in bits) of the SSH private key.
+    returned: changed or success
+    type: int
+    sample: 4096
+fingerprints:
+    description: Fingerprint(s) of newly created or deleted key(s)
+    return: changed and `return_fingerprints`==True
+    type: list[str]
+    sample: [ ABC123... ]
+'''
+
+from typing import Dict, Union
+
+import itertools
+import re
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.community.crypto.plugins.module_utils.gnupg.cli import GPGError
+from ansible_collections.community.crypto.plugins.plugin_utils.gnupg import GPGError
+
+def validate_params(params: Dict[str, Union[str, int]]) -> None:
+
+    if params['override'] and params['present'] and not (params['fingerprint'] or params['name'] or params['comment'] or params['email']):
+            raise GPGError, 'To override existing keys, please provide any combination of the `fingerprint`, `name`, `comment`, and `email` parameters.'
+    keys = ['key']
+    if params['subkey_type']:
+        keys.append('subkey')
+    for key in keys:
+        if params[f'{key}_type'] == 'EDDSA':
+            if not params[f'{key}_usage']: params[f'{key}_usage'] = ['sign', 'auth']
+            elif params[f'{key}_usage'] not in list(itertools.combinations(['sign', 'auth'])):
+                raise GPGError, f'Invalid {key}_usage for {params[f"{key}_type"]} {key}.'
+            if not params[f'{key}_curve'] or params[f'{key}_curve'] == 'cv25519':
+                params[f'{key}_curve'] = 'ed25519'
+            elif params[f'{key}_curve'] != 'cv25519':
+                raise GPGError, f'Invalid {key}_curve for {params[f"{key}_type"]} {key}.'  
+        elif params[f'{key}_type'] == 'ECDH':
+            if not params[f'{key}_usage']: params[f'{key}_usage'] = ['encrypt']
+            elif params[f'{key}_usage'] != ['encrypt']:
+                raise GPGError, f'Invalid {key}_usage for {params[f"{key}_type"]} {key}.'
+            if not params[f'{key}_curve']: params[f'{key}_curve'] = 'cv25519'
+            elif params[f'{key}_curve'] != 'cv25519':
+                raise GPGError, f'Invalid {key}_curve for {params[f"{key}_type"]} {key}.'  
+        elif params[f'{key}_type'] == 'ECDSA':
+            if not params[f'{key}_usage']: params[f'{key}_usage'] = ['sign', 'auth']
+            elif params[f'{key}_usage'] not in list(itertools.combinations(['sign', 'auth'])):
+                raise GPGError, f'Invalid {key}_usage for {params[f"{key}_type"]} {key}.'
+            if not params[f'{key}_curve']: params[f'{key}_curve'] = 'brainpoolp521r1'
+            elif params[f'{key}_curve'] not in ['nistp256', 'nistp384', 'nistp521', 'brainpoolP256r1', 'brainpoolP384r1', 'brainpoolP512r1', 'secp256k1']:
+                raise GPGError, f'Invalid {key}_curve for {params[f"{key}_type"]} {key}.'  
+        elif params[f'{key}_type'] == 'RSA':
+            if not params[f'{key}_usage']: params = ['ecrypt', 'sign', 'auth']
+            elif not params[f'{key}_usage'] not in list(itertools.combinatios(['ecrypt', 'sign', 'auth'])):
+                raise GPGError, f'Invalid {key}_usage for {params[f"{key}_type"]} {key}.'
+            if not params[f'{key}_length']: params[f'{key}_length'] = 3072
+            elif not 1024 <= params[f'{key}_length'] < 4096:
+                params[f'{key}_length'] = min(max(params[f'{key}_length'], 1024), 4096)
+        elif params[f'{key}_type'] == 'DSA':
+            if not params[f'{key}_usage']: params[f'{key}_usage'] = ['sign', 'auth']
+            elif params[f'{key}_usage'] not in list(itertools.combinations(['sign', 'auth'])):
+                raise GPGError, f'Invalid {key}_usage for {params[f"{key}_type"]} {key}.'
+            if not params[f'{key}_length']: params[f'{key}_length'] = 2048
+            elif not 768 <= params[f'{key}_length'] < 3072:    
+                params[f'{key}_length'] = min(max(params[f'{key}_length'], 768), 3072)
+        elif params[f'{key}_type'] == 'ELG':
+            if params[f'{key}_type'] == params['key_type']:
+                raise GPGError, f'Invalid algorithm for {key}_type parameter.'
+            if not params[f'{key}_usage']: params[f'{key}_usage'] = ['encrypt']
+            elif params[f'{key}_usage'] != ['encrypt']:
+                raise GPGError, f'Invalid {key}_iusage for {params[f"{key}_type"]} {key}.'
+            if not params[f'{key}_length']: params[f'{key}_length'] = 3072
+            elif not 1024 <= params[f'{key}_length'] < 4096:
+                params[f'{key}_length'] = min(max(params[f'{key}_length'], 1024), 4096)
+
+def list_matching_keys(name, comment, email, fingerprint):
+    user_id = ""
+    if params['name']:
+        user_id += f'{params["name"]} '
+    if params['comment']:
+        user_id += f'({params["comment"]}) '
+    if params['email']:
+        user_id += f'<{params["email"]}>'
+    if user_id:
+        user_id = f'"{user_id.strip()}"'
+
+    if user_id or fingerprints:
+        _, stdout, _ = gpg_runner.run_command(['gpg', '--batch', '--list-secret-keys', f'{*fingerprints if fingerprints else user_id}'])
+        lines = stdout.split('\n')
+        matching_keys = [line.strip() for line in lines if line.strip().isalnum()]
+        return matching_keys
+    return []
+
+def delete_keypair(
+    gpg_runner: PluginGPGRunner,
+    matching_keys: List[str],
+    check_mode: bool = False
+) -> Dict[str, Union[str, int]]:
+    if matching_keys:
+        gpg_runner.run_command([
+            f'{"dry-run" if check_mode else ""}',
+            '--batch',
+            '--yes',
+            '--delete-secret-and-public-key',
+            *matching_key
+        ], check_rc=True)
+        if params['return_fingerprints']:
+            return dict(changed=True, fingerprints=matching_keys)
+        return dict(changed=True, fingerprints=[])
+    return dict(changed=False, fingerprints=[])
+
+def generate_keypair(
+    gpg_runner: PluginGPGRunner,
+    params: Dict[str, Union[str, int, bool, List[str]]],
+    matching_keys,
+    check_mode: bool = False
+) -> Dict[str, Union[bool, str]]:
+    if matching_keys:
+        if params['return_fingerprints']:
+            return dict(changed=False, fingerprints=matching_keys)
+        return dict(change=False, fingerprints=[])
+
+    parameters = f"""<<EOF
+        Key-Type: {params['key_type']}
+        Key-Length: {params['key_type']}
+        Key-Curve: {params['key_curve']}
+        {f'''
+        Subkey-Type: {params["subkey_type"]}
+        Subkey-Length: {params["subkey_type"]}
+        Subkey-Curve: {params["subkey_curve"]}
+        ''' if params['subkey_type'] else ''}
+        Expire-Date: {params['expire_date']}
+        {f'Name-Real: {params["name"]}' if params['name'] else ''}
+        {f'Name-Comment: {params["comment"]}' if params['comment'] else ''}
+        {f'Name-Email: {params["email"]}' if params['email'] else ''}
+        {f'Passphrase: {params["passphrase"]}' if params['passphrase'] else '%no-protection'}
+        {f'Keyserver: {params["keyserver"]}' if params['keyserver'] else ''}
+        {'%transient-key' if params['transient_key'] else ''}
+        {'%dry-run' if check_mode or not params['force'] else ''}
+        %commit
+        EOF
+        """
+
+    _, stdout, _ = gpg_runner.run_command([
+        f'{"dry-run" if check_mode else ""}',
+        '--batch',
+        '--log-file',
+        '/dev/stdout',
+        '--gen-key',
+        f'{parameters}'
+    ])
+
+    if params['return_fingerprints']:
+        fingerprints = []
+        fingerprint = re.search(r"([a-zA-Z0-9]*)\.rev", stdout)
+        if fingerprint:
+            fingerprints.append(fingerprint)
+        return dict(changed=True, fingerprints=fingerprints)
+    return dict(changed=True, fingerprints=[])
+
+def run_module(params: Dict[str, Union[str, int, bool, List[str]]], check_mode: bool = False):
+    validate_params(params)
+    gpg_runner = PluginGPGRunner()
+    matching_keys = list_matching_keys(
+        params["name"],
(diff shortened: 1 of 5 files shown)


plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+    - "This module allows one to generate or delete OpenSSH private and public keys using GnuPG (gpg)."
+requirements:
+    - gpg >= 2.1
+extends_documentation_fragment:
+    - ansible.builtin.files
+    - community.crypto.attributes
+    - community.crypto.attributes.files
+attributes:
+    check_mode:
+        support: full
```
felixfontein, line 26:
```suggestion
        support: full
    diff_mode:
        support: none
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: gpg_keypair
+author: "Austin Lucas Lake (@austinlucaslake)"
+short_description: Generate or delete GPG private and public keys
```
felixfontein, line 15:
```suggestion
short_description: Generate or delete GPG private and public keys
version_added: 2.20.0
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+            - Whether the private and public keys should exist or not, taking action if the state is different from what is stated.
+        type: str
+        default: present
+        choices: [ present, absent ]
+    key_type:
+        description:
+            - "Specifies the type of key to create. By default this is V(EDDSSA) which must be used with Curve25519.
+              Supported key types are V(RSA), V(DSA), V(ECDSA), V(EDDSA), and V(ECDH)."
+        type: str
+        default: EDDSA
```
felixfontein, line 39: How about not having a default here? That way the default used by `gpg` would be used.

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+        type: str
+        choices: ['cv25519', 'nistp256', 'nistp384', 'nistp521', 'brainpoolP256r1', 'brainpoolP384r1', 'brainpoolP512r1', 'secp256k1']
+    key_usage:
+        description:
+            - Specifies usage(s) for key.
+            - Support usages are V(encrypt), V(sign), V(auth), V(cert).
+            - V(cert) is given to all primary keys regardess, however can be used to only give V(vert) usage to a key.
+            - If not usage is specified, the valid usages for the given key type with be assigned.
+            - If O(state) is V(absent), this parameter is ignored. 
+        type: list[str]
```
felixfontein, line 66:
```suggestion
        type: list
        elements: str
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+        description:
+            - Similar to O(key_length).
+            - For ELG keys, the minimum is V(1024), the maximum is V(4096), and the default is V(3072)."
+        type: int
+    subkey_curve:
+        description:
+            - "Similar to O(key_curve)"
+        type: str
+        choices: ['cv25519', 'nistp256', 'nistp384', 'nistp521', 'brainpoolP256r1', 'brainpoolP384r1', 'brainpoolP512r1', 'secp256k1']
+    key_usage:
```
felixfontein, line 84: Should this be `subkey_usage`?

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+        description:
+            - Similar to O(key_usage), but does not support V(cert).
+        type: list[str]
+        choices: ['encrypt', 'sign', 'auth', 'cert']
```
felixfontein, line 88: About all the `subkey_*` options: how about having a `subkey` option which is a list of structured dicts with keys like `type`, `length`, `curve`, `key_usage`, ...? Then you can also specify multiple subkeys.

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+    passphrase:
+        description:
+            - Passphrase used to decrypt an existing private key or encrypt a newly generated private key.
+            - If O(state) is V(absent), this parameter is ignored. 
+        type: str
+    fingerprints:
+        description:
+            - Specifies keys to match against.
+            - Provided fingerprints will take priority over user-id "V(name) (V(comment)) <V(email)>".
+            - If O(state) is V(absent), keys with the provided fingerprints will be deleted if found
```
felixfontein, line 110:
```suggestion
            - If O(state) is V(absent), keys with the provided fingerprints will be deleted if found.
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+            - If O(state) is V(absent), this parameter will be ignored.
+        type: str
+    transient_key:
+        description:
+            - Allows key generation to use a faster, but less secure random number generator.
+        type: bool
+        default: False
+    return_fingerprints:
+        description:
+            - Allows for the return of fingerprint(s) for newly created or deleted keys(s)
```
felixfontein, line 124:
```suggestion
            - Allows for the return of fingerprint(s) for newly created or deleted keys(s).
```
But why not simply always return the fingerprints?

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+  community.crypto.gpg_keypair:
+    state: abscent
+    fingerprints:
+      - ABC123...
+
+'''
+
+RETURN = '''
+size:
+    description: Size (in bits) of the SSH private key.
```
felixfontein, line 184: SSH → GPG? :)

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+    returned: changed or success
+    type: int
+    sample: 4096
+fingerprints:
+    description: Fingerprint(s) of newly created or deleted key(s)
+    return: changed and `return_fingerprints`==True
```
felixfontein, line 190:
```suggestion
    return: changed and O(return_fingerprints=true)
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+import re
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.community.crypto.plugins.module_utils.gnupg.cli import GPGError
+from ansible_collections.community.crypto.plugins.plugin_utils.gnupg import GPGError
+
+def validate_params(params: Dict[str, Union[str, int]]) -> None:
+
+    if params['override'] and params['present'] and not (params['fingerprint'] or params['name'] or params['comment'] or params['email']):
+            raise GPGError, 'To override existing keys, please provide any combination of the `fingerprint`, `name`, `comment`, and `email` parameters.'
```
felixfontein, line 207:
```suggestion
            raise GPGError('To override existing keys, please provide any combination of the `fingerprint`, `name`, `comment`, and `email` parameters.')
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+from typing import Dict, Union
+
+import itertools
+import re
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.community.crypto.plugins.module_utils.gnupg.cli import GPGError
+from ansible_collections.community.crypto.plugins.plugin_utils.gnupg import GPGError
+
+def validate_params(params: Dict[str, Union[str, int]]) -> None:
```
felixfontein, line 204: Please don't use type hints. These don't work with Python 2.

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+
+    if params['override'] and params['present'] and not (params['fingerprint'] or params['name'] or params['comment'] or params['email']):
+            raise GPGError, 'To override existing keys, please provide any combination of the `fingerprint`, `name`, `comment`, and `email` parameters.'
+    keys = ['key']
+    if params['subkey_type']:
+        keys.append('subkey')
+    for key in keys:
+        if params[f'{key}_type'] == 'EDDSA':
+            if not params[f'{key}_usage']: params[f'{key}_usage'] = ['sign', 'auth']
```
felixfontein, line 213: `f` strings don't work with Python 2 either. You need to use `%` or `str.format()`.

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+def run_module(params: Dict[str, Union[str, int, bool, List[str]]], check_mode: bool = False):
+    validate_params(params)
+    gpg_runner = PluginGPGRunner()
+    matching_keys = list_matching_keys(
+        params["name"],
+        params["comment"],
+        params["email"],
+        params["fingerprints"]
+    )
+    result = generate_keypair(gpg_runner, params, matching_keys, check_mode) if params['state'] == 'present' else delete_keypair(gpg_runner, matching_keys, check_mode)
```
felixfontein, line 352: It's easier to read if you split this up:
```suggestion
    if params['state'] = present:
        result = generate_keypair(gpg_runner, params, matching_keys, check_mode)
    else:
        result = delete_keypair(gpg_runner, matching_keys, check_mode)
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+        {f'Name-Email: {params["email"]}' if params['email'] else ''}
+        {f'Passphrase: {params["passphrase"]}' if params['passphrase'] else '%no-protection'}
+        {f'Keyserver: {params["keyserver"]}' if params['keyserver'] else ''}
+        {'%transient-key' if params['transient_key'] else ''}
+        {'%dry-run' if check_mode or not params['force'] else ''}
+        %commit
+        EOF
+        """
+
+    _, stdout, _ = gpg_runner.run_command([
```
felixfontein, line 326:
```suggestion
    dummy, stdout, dummy2 = gpg_runner.run_command([
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+    matching_keys: List[str],
+    check_mode: bool = False
+) -> Dict[str, Union[str, int]]:
+    if matching_keys:
+        gpg_runner.run_command([
+            f'{"dry-run" if check_mode else ""}',
+            '--batch',
+            '--yes',
+            '--delete-secret-and-public-key',
+            *matching_key
```
felixfontein, line 287:
```suggestion
            *matching_keys
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,390 @@
...
+    if params['name']:
+        user_id += f'{params["name"]} '
+    if params['comment']:
+        user_id += f'({params["comment"]}) '
+    if params['email']:
+        user_id += f'<{params["email"]}>'
+    if user_id:
+        user_id = f'"{user_id.strip()}"'
+
+    if user_id or fingerprints:
```
felixfontein, line 269: The argument is called `fingerprint` (singular).

felixfontein (commented):
Thanks for your contribution! I've added some first comments below.

One thing that should be discussed is idempotence. I.e. what does happen if parameters like `key_type` do not match the existing key (if `state=present`).

Generally Ansible should ensure the state, i.e. if the existing key has the wrong type, size, key usage, ..., it should be modified to the module's parameters. For objects such as private keys (or disk partitions etc.), destructive operations are generally a bad idea, so it makes sense to make that configurable, but other changes (like key usage) can be made without cycling the private key. For the private key handling modules in this collection, we usually make this behavior configurable (consider for example the `regenerate` option of `community.crypto.openssl_privatekey`).

Do you already have plans (after all this PR is marked as a draft)?

plugins/modules/gpg_keypair.py
austinlucaslake: Yes, it should be subkey_usage. Thanks for catching that.

plugins/modules/gpg_keypair.py
austinlucaslake: I'll need do some refactor as I used quite a few f strings haha. Refactoring will be commited later today.

austinlucaslake:
> Thanks for your contribution! I've added some first comments below.
> 
> One thing that should be discussed is idempotence. I.e. what does happen if parameters like `key_type` do not match the existing key (if `state=present`).
> 
> Generally Ansible should ensure the state, i.e. if the existing key has the wrong type, size, key usage, ..., it should be modified to the module's parameters. For objects such as private keys (or disk partitions etc.), destructive operations are generally a bad idea, so it makes sense to make that configurable, but other changes (like key usage) can be made without cycling the private key. For the private key handling modules in this collection, we usually make this behavior configurable (consider for example the `regenerate` option of `community.crypto.openssl_privatekey`).
> 
> Do you already have plans (after all this PR is marked as a draft)?

Thanks for the comments! Regeneration as you described, would be great. However there is a problem when matching against specific key types. The `key_type` parameter can't always be extracted when listing keys. For clarification, please take a look at the example outputs below when listing keys...

This is the output for an ECDSA key using the secp256k1 curve:
```
sec   secp256k1/44A9910642E57A60 2024-05-01 [SCA]
      CF68D75C4321667AE7E815F244A9910642E57A60
uid                 [ultimate] name (comment) <name@email.com>
```
This the output for an ECDSA key using the brainpoolP512r1 curve:
```
sec   brainpoolP512r1/5A6CA22A7DA2FE9C 2024-05-01 [SCA]
      F6892C47B20B59ACB652F3B95A6CA22A7DA2FE9C
uid                 [ultimate] name (comment) <name@email.com>
```
This is the output for an ECDSA key with an ECDH subkey, both of which use brainpoolP521r1 curve:
```
sec   brainpoolP512r1/AA5C22A75FD5B23D 2024-05-01 [SCA]
      E2280C64B089AEBCD7CE4EBEAA5C22A75FD5B23D
uid                 [ultimate] name (comment) <name@email.com>
ssb   brainpoolP512r1/836D4694DAF4DE3B 2024-05-01 [E]
```

This problem really only applies for when trying to decipher between ECDSA and ECDH keys. RSA, DSA, and ELG keys all include their key_type tags in the output along with the bit length. And EDDSA is output as ed25519 as thats the only curve its compatible with. Now that I think about it, it may be possible to differentiate between ECDSA and ECDH if I factor in the key usage as well. I'll see what I can do in that regard in the next couple of days.

plugins/modules/gpg_keypair.py
austinlucaslake: Creating GPG keys in batch mode (i.e. without user interaction) doesn't support multiple keys as is, but I could do some modifications to the code do get around that. I'll give that suggestion a try once I address some of the other things you mentioned.

plugins/modules/gpg_keypair.py
austinlucaslake: Thought I deleted the size return value entirely. Whoops

plugins/modules/gpg_keypair.py
austinlucaslake: I'm not 100% sure where the community would stand on this, but I had wanted to give users the option to keep the fingerprints hidden.

felixfontein:
Did you try to look at the `--with-colons` output?

austinlucaslake:
> Did you try to look at the `--with-colons` output?

Just took a look at that now. That should work. From what I see from the output, the OpenPGP algorithm id numbers can be parsed. I can convert those into the corresponding values for the `key_type` or `subkey_type` parameters input by users of the module.

plugins/modules/gpg_keypair.py
austinlucaslake: So in order to implement this, I would need to use loop and repeat `gpg --batch --quick-add-key` n-times to add an arbitrary number of subkeys. The command `--quick-add-key` accepts an `algo` parameter when can be something like `rsa`, `dsa`, or `rsa4096` and the like if you also wish to specify the number of bits. For ECC, its a little different. ECC keys require you to put the desired curve into the `algo` parameter instead of a key type such as `ecdsa`, `eddsa`, `ecdh`. Here's where a problem comes up: it seems you are unable to produce ECDSA subkeys using this method. Anytime you enter one of its supported curves (i.e. `nistp256`, `nistp384`, `nistp521`, `brainpoolP256r1`, `brainpoolP384r1`, `brainpoolP512r1`, `secp256k1`). This same list of curved is also supported but ECDH, and gpg will unfortunately defaults to it anytime you entire one of aforementioned curves into the `algo` parameter.

TLDR:
A tradeoff arises that's as follows: do you want an ECDSA subkey or do you want to have an arbitrary number of subkeys?

plugins/modules/gpg_keypair.py
austinlucaslake: Ended up removing the return_fingerprints since public keys naturally show the fingerprints anyways.

plugins/modules/gpg_keypair.py
austinlucaslake: Marking this as resolved for now. I made changes to allow for multiple subkeys. The caveat mentioned above is present, so this conversion can be reopened and discussed more in need be.

austinlucaslake:
Will be working on conditional regeneration over the next couple of days. Should be ready for review once that is fleshed out.

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,477 @@
...
+            - List of subkeys with their own respective key types, lengths, curves, and usages.
+            - Similar to O(key_type), O(key_length), O(key_curve), and (key_usage).
+            - Supports ECDH and ELG keys.
+            - For both ECDH and ELG keys, the only supported usage is V(encrypt).
+            - For ECDH keys, the default curve is V(brainpoolP512r1).
+            - ECDH keys also support the V(cv25519) curve.
+            - For ELG keys, the minimum length is V(1024) bits, the maximum length is V(4096) bits, and the default length is V(3072) bits.
+        type: list
+        elements: dict
+        options:
```
felixfontein, line 77: In `DOCUMENTATION` it's `suboptions`, not `options`:
```suggestion
        suboptions:
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,477 @@
...
+    type: list
+    elements: str
+    sample: [ ABC123... ]
+'''
+
+import itertools
+import re
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.community.crypto.plugins.plugin_utils.gnupg.py import PluginGPGRunner
```
felixfontein, line 179: First, the `.py` shouldn't be there, and second, this import does not work for modules, only for non-module plugins. Modules can only import things from module_utils.

For modules you'll need to create a new `GPGRunner` impementation (in `module_utils/gnupg/`, say a new file `module_runner.py`) that uses `AnsibleModule.run_command()`.

austinlucaslake: Quick question on your PluginGPGRunner code. I notice that its a child class that dervices from the abstract bass class of GPGRunner, which you also wrote. What's the reasoning for doing that? Why not just put the functionality PluginGPGRunner in GPGRunner.

felixfontein: `PluginGPGRunner` is only acceptable for plugins, not for modules, since it doesn't (and cannot) use `AnsibleModule.run_command`.

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+short_description: Generate or delete GPG private and public keys
+version_added: 2.20.0
+description:
+    - "This module allows one to generate or delete GPG private and public keys using GnuPG (gpg)."
+requirements:
+    - gpg >= 2.1
+attributes:
+    check_mode:
+        description:
+            - Can run in check_mode and return changed status prediction without modifying target.
```
felixfontein, line 24: Either change `description` to `details`, or (better!) simply remove it. The default description is there for a reason.

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+        support: full
+    diff_mode:
+        description:
+            - Will return details on what has changed (or possibly needs changing in check_mode), when in diff mode.
```
felixfontein, line 28: Same here.

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+DOCUMENTATION = '''
+---
+module: gpg_keypair
+author: "Austin Lucas Lake (@austinlucaslake)"
+short_description: Generate or delete GPG private and public keys
+version_added: 2.20.0
+description:
+    - "This module allows one to generate or delete GPG private and public keys using GnuPG (gpg)."
+requirements:
+    - gpg >= 2.1
```
felixfontein, line 20: You need the default `attributes` docs fragment:
```suggestion
    - gpg >= 2.1
extends_documentation_fragment:
  - community.crypto.attributes
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+            - Whether the private and public keys should exist or not, taking action if the state is different from what is stated.
+        type: str
+        default: present
+        choices: [ present, absent ]
+    key_type:
+        description:
+            - Specifies the type of key to create.
+        type: str
+        choices: [ 'RSA', 'DSA', 'ECDSA', 'EDDSA' ]
+    key_length:
```
felixfontein, line 42: In the collection, the term `key size` is usually used instead of `key length`. GPG seems to use both terms (at least on the man page), so I would prefer to use `key_size` here (if you want you can add `key_length` as an alias).

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+        description:
+            - For non-ECC keys, this specifies the number of bits in the key to create.
+            - For RSA keys, the minimum is V(1024), the maximum is V(4096), and the default is V(3072).
+            - For DSA keys, the minimum is V(768), the maximum is V(3072), and the default is V(2048).
+            - As per gpg's behavior, values below the allowed ranges will be set to the respective defaults, and values above will saturate at the maximum.
```
felixfontein, line 47: Generally if you refer to the program name, use `GPG` (capitalized). Only if you explicitly refer to the GPG executable, use `C(gpg)` (lower-case and code-style).
```suggestion
            - As per GPG's behavior, values below the allowed ranges will be set to the respective defaults, and values above will saturate at the maximum.
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+            - If O(key_curve=ed25519) is only supported if O(key_type=EDDSA).
+            - This is required if O(key_type=ECDSA) or O(key_type=EDDSA) and it is ignored if O(key_type=RSA) or O(key_type=DSA).
+        type: str
+        choices: [ 'nistp256', 'nistp384', 'nistp521', 'brainpoolP256r1', 'brainpoolP384r1', 'brainpoolP512r1', 'secp256k1', 'ed25519' ]
+    key_usage:
+        description:
+            - Specifies usage(s) for key.
+            - V(cert) is given to all primary keys regardess, however can be used to only give V(vert) usage to a key.
+            - If not usage is specified, all of valid usages for the given key type are assigned.
+            - O(key_usage=encr) is only supported is O(key_type=RSA).
```
felixfontein, line 62:
```suggestion
            - O(key_usage=encr) is only supported if O(key_type=RSA).
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+            subkey_length:
+                description:
+                    - Similar to O(key_length).
+                    - For ELG subkeys, the minimum length is V(1024) bits, the maximum length is V(4096) bits, and the default length is V(3072) bits.
+                type: int
+            subkey_curve:
+                description:
+                    - Similar to O(key_curve).
+                    - V(cv25519) is supported if subkey_type is V(ECDH).
+                    - This is required if subkey_type is V(ECDSA), V(EDDSA), or V(ECDH) and it is ignored if subkey_type is V(RSA), V(DSA), or V(ELG).
```
felixfontein, line 89:
```suggestion
                    - This is required if subkey_type is V(ECDSA), V(EDDSA), or V(ECDH) and it is ignored if O(subkeys[].subkey_type) is V(RSA), V(DSA), or V(ELG).
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+                type: str
+                choices: ['nistp256', 'nistp384', 'nistp521', 'brainpoolP256r1', 'brainpoolP384r1', 'brainpoolP512r1', 'secp256k1', 'ed25519', 'cv25519']
+            subkey_usage:
+                description:
+                    - Similar to O(key_usage).
+                    - V(encr) is supported if subkey_type is V(RSA), V(ECDH), or V(ELG).
```
felixfontein, line 95:
```suggestion
                    - V(encr) is supported if O(subkeys[].subkey_type) is V(RSA), V(ECDH), or V(ELG).
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+        default: []
+        choices: [ 'encr', 'sign', 'auth', 'cert' ]
+    subkeys:
+        description:
+            - List of subkeys with their own respective key types, lengths, curves, and usages.
+        type: list
+        elements: dict
+        default: []
+        suboptions:
+            subkey_type:
```
felixfontein, line 74: How about using
```suggestion
            key_type:
```
here (similar below?). That this is already a subkey is clear from this being part of `subkeys`.

Using the same keys as for the primary key makes it easy to copy over the values - you only have to change the indentation (except for the `-` before the first key for the subkey).

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+                choices: [ 'encr', 'sign', 'auth' ]
+    expire_date:
+        description:
+            - Sets the expire date for the key.
+            - If O(expire_date=0), the key never expires.
+            - If O(expire_date=<n>), the key expires in V(n) days.
+            - If O(expire_date=<n>w), the key expires in V(n) weeks.
+            - If O(expire_date=<n>m), the key expires in V(n) months.
+            - If O(expire_date=<n>y), the key expires in V(n) years.
+            - If left unspecified, any created GPG keys never expire.
```
felixfontein, line 109: Is it also possible to allow providing an explicit date (say in ISO format, `yyyy-mm-dd`)? Then this option would be properly idempotent, since the expiration date would be an absolute value and not depend on today's date.

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+        type: str
+    name:
+        description:
+            - Specifies a name for the key's user id.
```
felixfontein, line 113:
```suggestion
            - Specifies a name for the key's user ID.
```
(https://docs.ansible.com/ansible/devel/dev_guide/style_guide/spelling_word_choice.html#acronyms)

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+        description:
+            - Specifies a comment for the key's user id.
+        type: str
+    email:
+        description:
+            - Specifies an email for the key's user id.
+        type: str
+    passphrase:
+        description:
+            - Passphrase used to decrypt an existing private key or encr a newly generated private key.
```
felixfontein, line 125:
```suggestion
            - Passphrase used to decrypt an existing private key or encrypt a newly generated private key.
```

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+        type: str
+    fingerprints:
+        description:
+            - Specifies keys to match against.
+        type: list
+        elements: str
+        default: []
+    force:
+        description:
+            - If O(force=True), key generation is executed using the module's options, even a matching key is found.
```
felixfontein, line 135:
```suggestion
            - If O(force=true), key generation is executed using the module's options, even a matching key is found.
```
Always use `true` and `false` for booleans outside Python code. (Documentation is not considered Python code, since it's YAML.)

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+            if subkey['subkey_length']:
+                algo += str(subkey['subkey_length'])
+        elif subkey['subkey_curve']:
+            algo = subkey['subkey_curve']
+        else:
+            algo = 'default'
+
+        module.run_command(
+            [
+                '--dry-run' if check_mode else '',
```
felixfontein, line 347: I don't think providing an empty argument is a good idea. If you want to skip it, use `None`, since `AnsibleModule.run_command()` will simply omit `None` values (as opposed to empty strings).

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+            ['--list-secret-keys', '--with-colons', fingerprint],
+            executable=get_bin_path('gpg'),
+            check_rc=True
+        )
+
+        subkey_count = 0
+        uid_present = not bool(uid)
+        for line in stdout.splitlines():
+            if line[:3] == 'sec':
+                primary_key = re.search(r'.+:([0-9]+):([0-9]+):[0-9A-Z]+:[0-9]+:[0-9]+::u:+([a-z]+)[A-Z]+:+\+:+([0-9a-zA-Z]+)::0:', line)
```
felixfontein, line 399: Why not simply split by `:`, and then access the fields?

plugins/modules/gpg_keypair.py
```diff
@@ -0,0 +1,507 @@
...
+                    subkey_type=dict(type='str', choices=key_types),
+                    subkey_length=dict(type='int', no_log=False),
+                    subkey_curve=dict(type='str', choices=key_curves),
+                    subkey_usage=dict(type='list', elements='str', default=[], choices=key_usages[:-1])
+                ),
+                required_if=[
+                    ['subkey_type', 'ECDSA', ['subkey_curve']],
+                    ['subkey_type', 'EDDSA', ['subkey_curve']],
+                    ['subkey_type', 'ECDH', ['subkey_curve']]
+                ]
```
felixfontein, line 480: Generally it's a good idea to always use trailing commas. This will reduce the size of diffs when this is ever changed.
```suggestion
                    ['subkey_type', 'ECDH', ['subkey_curve']],
                ],
```

plugins/modules/gpg_keypair.py
austinlucaslake: I added the description cause some of the tests were throwing errors without it.

plugins/modules/gpg_keypair.py
austinlucaslake: Same as above.

plugins/modules/gpg_keypair.py
austinlucaslake: Much simpiler than I was thinking haha. Seems like a good change.

plugins/modules/gpg_keypair.py
austinlucaslake: GPG supports this, but I'm still thinking of a way to verify date formats nicely. For example Python 3.7 and onwards has `datetime.fromisoformat()` which could be wrapped in a try/except block in case of invalid date formats. Is there such a function for Python 2.6/2.7 for people running this on older versions?

plugins/modules/gpg_keypair.py
austinlucaslake: Sounds good. Did that originally cause of name collision in documentation when using O(), but one of your other eliminates that issue.

plugins/modules/gpg_keypair.py
austinlucaslake: Marking as resolved. Seems like the `dateutil` package will satisfy this functionality.

felixfontein: Please do not use dateutil, it's nt part of the Python standard library. You can use `datetime.datetime.strptime()` with an appropriate format string (`%Y-%m-%d`).

PR #757, opened 2024-05-12
ACME: add new modules acme_certificate_order_create, acme_certificate_order_finalize, acme_certificate_order_info, acme_certificate_order_validate for working with ACME orders

felixfontein:
##### SUMMARY
Adds a set of new modules to work with ACME orders. These modules only work with ACME v2. (We already deprecated ACME v1 support anyway.)

This aims to eventually replaces the acme_certificate module, replacing it by a set of modules that are more flexible and easier to use.

These modules are aimed for role authors (such as myself :) ), not really for end-users.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
acme_certificate_order_create
acme_certificate_order_finalize
acme_certificate_order_info
acme_certificate_order_validate

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,10 @@ If you use the Ansible package and do not update collections independently, use 
   - acme_ari_info module
   - acme_certificate module
   - acme_certificate_deactivate_authz module
+  - acme_certificate_order_create module
+  - acme_certificate_order_finalize module
+  - acme_certificate_order_info module
+  - acme_certificate_order_validate module
   - acme_certificate_revoke module
   - acme_challenge_cert_helper module
   - acme_inspect module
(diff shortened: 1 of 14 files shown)


felixfontein:
I still have zero idea what is going wrong, since I simply cannot reproduce these errors locally. :-(

felixfontein:
One of my remaining guesses was that this is due to the VMs used to run the tests only have one CPU, and that is causing this strange behavior somehow. Well, I've run this in a 1 CPU VM myself on my machine, and there it also doesn't fail. I'm really at a loss here...

felixfontein:
I found a difference between CI and my local setup: https://github.com/ansible-collections/community.crypto/blob/main/tests/utils/shippable/generic.sh#L19

felixfontein:
Yay, finally CI is passing. That took some time :)

felixfontein:
I'm going to merge this once CI passes and push a PR to https://github.com/felixfontein/ansible-acme/ which uses the new modules.

felixfontein:
Ref: https://github.com/felixfontein/ansible-acme/pull/86

PR #777, opened 2024-07-05
certificate_complete_chain: add ability to identify ed25519 complete chains

gderber:
##### SUMMARY
This adds the ability to use the certificate_complete_chain module on ed25519 certificates.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
certificate_complete_chain

##### ADDITIONAL INFORMATION
Without this change, using this module with an ed25519 cert provides an error:
[WARNING]: Unknown public key type "<cryptography.hazmat.backends.openssl.ed25519._Ed25519PublicKey object at 0x7f5263b8bd90>"

Upon making the change, it works.  However I do not like it.  The reason I don't like it is because the "verify" method for class Ed25519PublicKey in the cryptography library is an abstractmethod with undefined functionality.  I could never find where the actual functionality is defined.  I don't really understand the cryptography library well, nor how this ansible module uses it.  Which means more work may be needed but I am having trouble identifying what that might be.

References:
- https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/primitives/asymmetric/ed25519.py

I am quite open to critics and pointers.

### DIFF
--- a/changelogs/fragments/777-add_ability_to_identify_ed25519_complete_chains.yml
+++ b/changelogs/fragments/777-add_ability_to_identify_ed25519_complete_chains.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - complete_chain - add ability to identify ed25519 and ed448 complete chains.
+  - certificate_complete_chain - add ability to identify ed25519 and ed448 complete chains (https://github.com/ansible-collections/community.crypto/pull/777).
--- a/changelogs/fragments/777-add_ability_to_identify_ed25519_complete_chains.yml
+++ b/changelogs/fragments/777-add_ability_to_identify_ed25519_complete_chains.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - complete_chain - add ability to identify ed25519 and ed448 complete chains.
--- a/changelogs/fragments/777-add_ability_to_identify_ed25519_complete_chains.yml
+++ b/changelogs/fragments/777-add_ability_to_identify_ed25519_complete_chains.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - certificate_complete_chain - add ability to identify ed25519 and ed448 complete chains (https://github.com/ansible-collections/community.crypto/pull/777).
+  - certificate_complete_chain - add ability to identify Ed25519 and Ed448 complete chains (https://github.com/ansible-collections/community.crypto/pull/777).
--- a/plugins/modules/certificate_complete_chain.py
+++ b/plugins/modules/certificate_complete_chain.py
@@ -150,6 +150,7 @@ try:
     import cryptography.hazmat.primitives.serialization
     import cryptography.hazmat.primitives.asymmetric.rsa
     import cryptography.hazmat.primitives.asymmetric.ec
+    import cryptography.hazmat.primitives.asymmetric.ed25519
     import cryptography.hazmat.primitives.asymmetric.padding
     import cryptography.hazmat.primitives.hashes
     import cryptography.hazmat.primitives.asymmetric.utils
@@ -195,6 +196,11 @@ def is_parent(module, cert, potential_parent):
                 cert.cert.signature,
                 cert.cert.tbs_certificate_bytes,
                 cryptography.hazmat.primitives.asymmetric.ec.ECDSA(cert.cert.signature_hash_algorithm),
+            )
+        elif isinstance(public_key, cryptography.hazmat.primitives.asymmetric.ed25519.Ed25519PublicKey):
+            public_key.verify(
+                cert.cert.signature,
+                cert.cert.tbs_certificate_bytes
             )
         else:
             # Unknown public key type
--- a/plugins/modules/certificate_complete_chain.py
+++ b/plugins/modules/certificate_complete_chain.py
@@ -141,8 +141,11 @@ from ansible_collections.community.crypto.plugins.module_utils.version import Lo
 from ansible_collections.community.crypto.plugins.module_utils.crypto.pem import (
     split_pem_list,
 )
+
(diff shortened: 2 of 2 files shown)


felixfontein:
The actual implementation is here: https://github.com/pyca/cryptography/blob/main/src/rust/src/backend/ed25519.rs#L123

felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

plugins/modules/certificate_complete_chain.py
```diff
@@ -196,6 +197,11 @@ def is_parent(module, cert, potential_parent):
                 cert.cert.tbs_certificate_bytes,
                 cryptography.hazmat.primitives.asymmetric.ec.ECDSA(cert.cert.signature_hash_algorithm),
             )
+        elif isinstance(public_key, cryptography.hazmat.primitives.asymmetric.ed25519.Ed25519PublicKey):
```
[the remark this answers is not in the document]
gderber: I was thinking about adding ed448 as well, since they are so similar. Wasn't sure if that'd be better as a separate PR or part of this one.

I'll add ed448 and the changelog fragment.  It might be a few days before I can test the ed448 addition.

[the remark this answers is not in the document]
felixfontein: Doing it in the same PR is fine, after all it's a pretty small change :) Thanks for doing this!

gderber:
I made the requested changes.  I did a force push to my repo because I wanted to change the author / email associated with the first commit.

changelogs/fragments/777-add_ability_to_identify_ed25519_complete_chains.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - complete_chain - add ability to identify ed25519 and ed448 complete chains.
```
felixfontein, line 2:
```suggestion
  - certificate_complete_chain - add ability to identify Ed25519 and Ed448 complete chains (https://github.com/ansible-collections/community.crypto/pull/777).
```

plugins/modules/certificate_complete_chain.py
```diff
@@ -141,6 +141,9 @@
 from ansible_collections.community.crypto.plugins.module_utils.crypto.pem import (
     split_pem_list,
 )
+from ansible_collections.community.crypto.plugins.module_utils.crypto.basic import (
+    CRYPTOGRAPHY_HAS_ED448_SIGN, CRYPTOGRAPHY_HAS_ED25519_SIGN)
+
```
felixfontein, line 146: To keep the style of the other imports:
```suggestion

from ansible_collections.community.crypto.plugins.module_utils.crypto.basic import (
    CRYPTOGRAPHY_HAS_ED448_SIGN,
    CRYPTOGRAPHY_HAS_ED25519_SIGN,
)
```

changelogs/fragments/777-add_ability_to_identify_ed25519_complete_chains.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - certificate_complete_chain - add ability to identify ed25519 and ed448 complete chains (https://github.com/ansible-collections/community.crypto/pull/777).
```
felixfontein, line 2: The algorithms are called `EdXXX`, with an upper-case `E`:
```suggestion
  - certificate_complete_chain - add ability to identify Ed25519 and Ed448 complete chains (https://github.com/ansible-collections/community.crypto/pull/777).
```

plugins/modules/certificate_complete_chain.py
```diff
@@ -142,6 +142,12 @@
     split_pem_list,
 )
 
+from ansible_collections.community.crypto.plugins.module_utils.crypto.basic import (
+    CRYPTOGRAPHY_HAS_ED448_SIGN,
+    CRYPTOGRAPHY_HAS_ED25519_SIGN,
+)
+
```
felixfontein, line 149: Please keep one line between import blocks.
```suggestion
```

plugins/modules/certificate_complete_chain.py
```diff
@@ -150,6 +156,8 @@
     import cryptography.hazmat.primitives.serialization
     import cryptography.hazmat.primitives.asymmetric.rsa
     import cryptography.hazmat.primitives.asymmetric.ec
+    import cryptography.hazmat.primitives.asymmetric.ed448
+    import cryptography.hazmat.primitives.asymmetric.ed25519
```
felixfontein, line 160: These two imports must not be there, otherwise this won't work with older cryptography versions:
```suggestion
```

gderber: I've got some testing to do then.

I thought I was getting errors without them, but the most recent run seems to be working okay without them.  I want to do further testing.

gderber: Well, I could never reproduce whatever bug caused me to add them in the first place.  It works without them.  Bug may have been elsewhere and I didn't diagnose it correctly.

felixfontein: The imports are only necessary if you want to make absolutely sure that these imports are available. But since these are optional features (checked with the feature flags - which are obtained by testing imports in plugins/module_utils/crypto/basic.py) there's no need to do the imports here, and in fact it breaks the module for older cryptography versions which don't have these imports.

felixfontein (approved):

felixfontein:
@gderber thanks for your contribution!

gderber:
You're welcome!  Thank you for your patience and guidance on this.

PR #779, opened 2024-07-06
Enable TLS/SSL CTX Options for the get_certificate Module

dlehrman:
##### SUMMARY
Adds an option for TLS/SSL CTX options to the get_certificate module. Accepts both string (e.g. OP_LEGACY_SERVER_CONNECT) and integer (e.g. 4) representations for options as some options can be enabled via integer but not string, depending on the Python and/or OpenSSL/LibreSSL versions.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
get_certificate

##### ADDITIONAL INFORMATION
I use Ansible to manage various device types (e.g. PDU's, BMC's, etc), some of which do not support secure renegotiation. For devices that do not support secure renegotiation, depending on the Ansible controller, the get_certificate module was experiencing an SSL failure, though I could still interact them with GUI web browsers (Chrome, Firefox).

Without manually enabling OP_LEGACY_SERVER_CONNECT (example endpoint uses 2048 bit cert, TLSv1.2 / AES128-GCM-SHA256, does not support secure renegotiation, cannot be changed):
```paste below
TASK [Get Certificate] ********************************************************************************************************************************
fatal: [REDACTED]: FAILED! => {"changed": false, "msg": "Failed to get cert from REDACTED, error: [SSL: UNSAFE_LEGACY_RENEGOTIATION_DISABLED] unsafe legacy renegotiation disabled (_ssl.c:1007)"}
```

After manually enabling OP_LEGACY_SERVER_CONNECT:
```
TASK [Get Certificate] ********************************************************************************************************************************
ok: [REDACTED]
```

Relevant task:
```yaml
    - name: "Get Certificate"
      community.crypto.get_certificate:
        host: "{{ ipv4_address}}"
        port: "{{ https_port}}"
        asn1_base64: true
        ciphers:
          - HIGH
        ssl_ctx_options: # defaults for my Ansible controller (details below), except for OP_LEGACY_SERVER_CONNECT
          - OP_ALL
          - OP_NO_SSLv3
          - OP_CIPHER_SERVER_PREFERENCE
          - OP_ENABLE_MIDDLEBOX_COMPAT
          - OP_NO_COMPRESSION
          - 4 # OP_LEGACY_SERVER_CONNECT
      register: certificate
```

Ansible controller properties:
- Ubuntu 22.04.4 LTS
- Ansible 2.17.0
- Python 3.10.6-1~22.04
- OpenSSL 3.0.2-0ubuntu1.15


plugins/modules/get_certificate.py
```diff
@@ -99,6 +99,14 @@
...
       type: bool
       version_added: 2.12.0
+    ssl_ctx_options:
+      description:
+        - SSL CTX options (SSL OP flags) to use for the request.
+        - See the L(List of SSL OP Flags,https://wiki.openssl.org/index.php/List_of_SSL_OP_Flags) for more details.
+        - The available SSL CTX options is dependent on the Python and OpenSSL/LibreSSL versions.
+      type: list
+      elements: [ str, int ]
+      version_added: tbd
```
dlehrman, line 109: Please advise what value to set for version_added.

felixfontein: Use 2.21.0, that will be the next minor release.

plugins/modules/get_certificate.py
```diff
@@ -99,6 +99,14 @@
         - The default value V(false) is B(deprecated) and will change to V(true) in community.crypto 3.0.0.
       type: bool
       version_added: 2.12.0
+    ssl_ctx_options:
```
felixfontein, line 102: Please use TLS instead of SSL. The protocol has been named TLS for 25 years by now.
```suggestion
    tls_ctx_options:
```

changelogs/fragments/779-add-ssl_ctx_options-option.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
+  - get_certificate - adds `ssl_ctx_options` option for specifying SSL CTX options (https://github.com/ansible-collections/community.crypto/pull/779).
```
felixfontein, line 3:
```suggestion
  - get_certificate - adds ``tls_ctx_options`` option for specifying SSL CTX options (https://github.com/ansible-collections/community.crypto/pull/779).
```

plugins/modules/get_certificate.py
```diff
@@ -99,6 +99,14 @@
...
+      description:
+        - SSL CTX options (SSL OP flags) to use for the request.
+        - See the L(List of SSL OP Flags,https://wiki.openssl.org/index.php/List_of_SSL_OP_Flags) for more details.
+        - The available SSL CTX options is dependent on the Python and OpenSSL/LibreSSL versions.
+      type: list
+      elements: [ str, int ]
```
felixfontein:
```suggestion
      elements: raw
```

plugins/modules/get_certificate.py
```diff
@@ -285,6 +311,7 @@ def main():
             starttls=dict(type='str', choices=['mysql']),
             ciphers=dict(type='list', elements='str'),
             asn1_base64=dict(type='bool'),
+            ssl_ctx_options=dict(type='list', default=None),
```
felixfontein, line 313: The default is always `None` if not specified. But you have to specify `elements`:
```suggestion
            ssl_ctx_options=dict(type='list', elements='raw'),
```

plugins/modules/get_certificate.py
```diff
@@ -381,6 +412,21 @@ def main():
...
+            if ssl_ctx_options is not None:
+                # Clear default options
+                ctx.options = 0
+
+                # For each item in the ssl_ctx_options list
+                for ssl_ctx_option in ssl_ctx_options:
+                    # If the item is a string
+                    if isinstance(ssl_ctx_option, str):
+                        # Get the int value for the option and add it to ctx options
+                        ctx.options |= getattr(ssl, ssl_ctx_option)
```
felixfontein, line 423: You need to handle the case that `getattr` returns `None` (or something that's not an integer).

plugins/modules/get_certificate.py
```diff
@@ -381,6 +412,21 @@ def main():
...
                 ctx.set_ciphers(ciphers_joined)
 
+            if ssl_ctx_options is not None:
+                # Clear default options
+                ctx.options = 0
+
+                # For each item in the ssl_ctx_options list
+                for ssl_ctx_option in ssl_ctx_options:
+                    # If the item is a string
+                    if isinstance(ssl_ctx_option, str):
```
felixfontein, line 421: Use `string_types` (`from ansible.module_utils.six import string_types`) instead of `str` here, and convert it to a native string with `to_native()` (`from ansible.module_utils.common.text.converters import to_native`) before using it.

plugins/modules/get_certificate.py
```diff
@@ -381,6 +412,21 @@ def main():
...
+                        # Get the int value for the option and add it to ctx options
+                        ctx.options |= getattr(ssl, ssl_ctx_option)
+                    # If the item is an integer
+                    elif isinstance(ssl_ctx_option, int):
+                        # Add the int value of the option to ctx options
+                        ctx.options |= ssl_ctx_option
```
felixfontein, line 427: You need an `else` to handle all other types.

felixfontein (commented):
Thanks for your contribution!

plugins/modules/get_certificate.py
```diff
@@ -99,6 +99,14 @@
         - The default value V(false) is B(deprecated) and will change to V(true) in community.crypto 3.0.0.
       type: bool
       version_added: 2.12.0
+    tls_ctx_options:
+      description:
+        - TLS CTX options (TLS/SSL OP flags) to use for the request.
```
felixfontein, line 104: I would write 'context' out at least once:
```suggestion
        - TLS context options (TLS/SSL OP flags) to use for the request.
```

plugins/modules/get_certificate.py
```diff
@@ -381,6 +413,39 @@ def main():
...
+                ctx.options = 0
+
+                # For each item in the tls_ctx_options list
+                for tls_ctx_option in tls_ctx_options:
+                    # If the item is a string_type
+                    if isinstance(tls_ctx_option, string_types):
+                        # Convert tls_ctx_option to a native string
+                        tls_ctx_option_str = to_native(tls_ctx_option)
+                        # Get the tls_ctx_option_str attribute from ssl
+                        tls_ctx_option_attr = getattr(ssl, tls_ctx_option_str)
```
felixfontein, line 427:
```suggestion
                        tls_ctx_option_attr = getattr(ssl, tls_ctx_option_str, None)
```
Otherwise you end up with an uncaught exception if a random string is provided that doesn't exist as an attribute of `ssl`.

plugins/modules/get_certificate.py
```diff
@@ -381,6 +413,39 @@ def main():
...
+                        # If tls_ctx_option_attr is not an integer
+                        else:
+                            module.fail_json(msg="Failed to determine the numeric value for {0}".format(tls_ctx_option_str))
+                    # If the item is an integer
+                    elif isinstance(tls_ctx_option, int):
+                        # Set tls_ctx_option_int to the item value
+                        tls_ctx_option_int = tls_ctx_option
+                    # If the item is not a string nor integer
+                    else:
+                        module.fail_json(msg="tls_ctx_options must be a string or integer")
```
felixfontein, line 441: Maybe provide the value here:
```suggestion
                        module.fail_json(msg="tls_ctx_options must be a string or integer, got {0!r}".format(tls_ctx_option))
```
That can be useful if an expression is used for this module option whose value isn't obvious.

plugins/modules/get_certificate.py
dlehrman: Random strings are actually handled as-is. I tried adding `OP_ALL_DOES_NOT_EXIST` to the options list without adding `, None` to `getattr` and this is the output:
```
TASK [Get Certificate] ********************************************************************************************************************************
fatal: [REDACTED]: FAILED! => {"changed": false, "msg": "Failed to get cert from REDACTED, error: module 'ssl' has no attribute 'OP_ALL_DOES_NOT_EXIST'"}
```

Do you still want this suggestion implemented?

felixfontein: The global try/catch is for catching errors in the actual certificate retrieval, not for argument handling. The error message also contains information on a proxy (if configured), which would be very confusing in this case. I think it's better to have clear error handling here which is not related ot failed certificate retrieval.

plugins/modules/get_certificate.py
dlehrman: Makes sense. Changing both instances in the description. Would you prefer `tls_ctx_options` to be changed to `tls_context_options` as well?

felixfontein: I think `ctx` in the option name is fine, as long as the documentation at least once mentions `context` :)

felixfontein (approved):

felixfontein:
@dlehrman thanks a lot for your contribution!

PR #784, opened 2024-07-09
get_certificate: add get_certificate_chain option

### LINKED ISSUE #568: Get a certificate chain with `crypto.get_certificate` module
##### SUMMARY

Currently, `crypto.get_certificate` module only returns a server certificate itself, but it is more useful to return all certificates sent by the server, as does the `-showcerts` option of `openssl s_client` command.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`crypto.get_certificate`

##### ADDITIONAL INFORMATION

For example, I would like to get CA certificate from a LDAP server and save it as a file in the client's trust store. If I don't already have the certificate saved on disk or in a variable, I should try to parse the command module output.

I don't think any additional options are needed here, just expanding the output.

Sample output:

```yaml
ok: [example] => {
    "my_retrieved_cert": {
        "cert": "-----BEGIN CERTIFICATE-----\nxSERVERxCERTIFICATEx==\n-----END CERTIFICATE-----\n",
        "intermediate": "-----BEGIN CERTIFICATE-----\nxINTERMEDIATExCERTIFICATEx==\n-----END CERTIFICATE-----\n",
        "root": "-----BEGIN CERTIFICATE-----\nxROOTxCERTIFICATEx==\n-----END CERTIFICATE-----\n",
        "changed": false,
        "expired": false,
...
```

felixfontein:
##### SUMMARY
Fixes #568.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
get_certificate

### DIFF
--- a/plugins/modules/get_certificate.py
+++ b/plugins/modules/get_certificate.py
@@ -509,13 +509,24 @@ def main():
                             return []
                         return [c.public_bytes(ssl._ssl.ENCODING_DER) for c in chain]
 
-                    ssl_obj = tls_sock._sslobj  # This is of type _ssl._SSLSocket
+                    ssl_obj = tls_sock._sslobj  # This is of type ssl._ssl._SSLSocket
                     verified_der_chain = _convert_chain(ssl_obj.get_verified_chain())
                     unverified_der_chain = _convert_chain(ssl_obj.get_unverified_chain())
                 else:
                     # This works with Python 3.13+
-                    verified_der_chain = tls_sock.get_verified_chain()
-                    unverified_der_chain = tls_sock.get_unverified_chain()
+
+                    # Unfortunately due to a bug (https://github.com/python/cpython/issues/118658) some early pre-releases of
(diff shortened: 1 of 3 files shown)


felixfontein:
It looks like the Python 3.13 API is somewhat broken, which makes the tests fail: https://github.com/python/cpython/issues/118658

There is a fix for it already: https://github.com/python/cpython/pull/118669

felixfontein:
It might be worth to include a similar hack as in https://github.com/pypa/pip/pull/12707/files#diff-5c235fa5de9ea9d116494690fc0e484c45501486f276649200c7e612c2477f88R285 to fix the Python 3.13 issues.

felixfontein:
ede3e54ba13cbf0d7f7e09dc388971c5b70775c5 makes tests pass with the current Python 3.13 pre-releases.

markuman (approved):

PR #803, opened 2024-10-01
lookup lowercase domain names when verifying authorizations to preven…

lspiehler:
…t failure when CSR has mixed-case names

##### SUMMARY
I received a CSR from a vendor that had uppercase characters in someof the requested domain names, which resulted in the following error when validating challenges with the acme_certificate module: "Found no authorization information for \"dns:My.FQDN.com\"!". The problem was caused by the authorizations in the returned challenge data having all domains (keys for the validation object) converted to lowercase, but when the keys were being looked up to make sure each is valid, the original case in the CSR was used.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
acme_certificate 

##### ADDITIONAL INFORMATION
Before change
```
TASK [ansible_acme : Complete challenge] ***************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Found no authorization information for \"dns:My.FQDN.com\"!", "other": {}}
```

After change, task completes successfully.

### DIFF
--- a/changelogs/fragments/803-fix-authorization-failure-with-mixed-case-sans.yml
+++ b/changelogs/fragments/803-fix-authorization-failure-with-mixed-case-sans.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - acme_certificate - fix authorization failure when CSR contains SANs with mixed case
+  - acme_certificate - fix authorization failure when CSR contains SANs with mixed case (https://github.com/ansible-collections/community.crypto/pull/803).
--- a/changelogs/fragments/803-fix-authorization-failure-with-mixed-case-sans.yml
+++ b/changelogs/fragments/803-fix-authorization-failure-with-mixed-case-sans.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - acme_certificate - fix authorization failure when CSR contains SANs with mixed case
--- a/plugins/module_utils/acme/challenges.py
+++ b/plugins/module_utils/acme/challenges.py
@@ -46,9 +46,11 @@ def create_key_authorization(client, token):
 def combine_identifier(identifier_type, identifier):
     return '{type}:{identifier}'.format(type=identifier_type, identifier=identifier)
 
+
 def normalize_combined_identifier(identifier):
     parts = split_identifier(identifier)
     return '{type}:{identifier}'.format(type=parts[0], identifier=parts[1].lower())
+
 
 def split_identifier(identifier):
     parts = identifier.split(':', 1)
--- a/plugins/module_utils/acme/challenges.py
+++ b/plugins/module_utils/acme/challenges.py
@@ -48,8 +48,10 @@ def combine_identifier(identifier_type, identifier):
 
 
 def normalize_combined_identifier(identifier):
-    parts = split_identifier(identifier)
-    return '{type}:{identifier}'.format(type=parts[0], identifier=parts[1].lower())
+    identifier_type, identifier = split_identifier(identifier)
+    # Normalize DNS names and IPs
+    identifier = identifier.lower()
+    return combine_identifier(identifier_type, identifier)
 
 
 def split_identifier(identifier):
--- a/plugins/module_utils/acme/challenges.py
+++ b/plugins/module_utils/acme/challenges.py
@@ -46,6 +46,9 @@ def create_key_authorization(client, token):
 def combine_identifier(identifier_type, identifier):
     return '{type}:{identifier}'.format(type=identifier_type, identifier=identifier)
 
+def normalize_combined_identifier(identifier):
+    parts = split_identifier(identifier)
+    return '{type}:{identifier}'.format(type=parts[0], identifier=parts[1].lower())
 
 def split_identifier(identifier):
     parts = identifier.split(':', 1)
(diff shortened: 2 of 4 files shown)


felixfontein (commented):
I think this change is incorrect, as this might only work with one specific CA, but could break behavior with other CAs.

This probably needs a more general solution that allows matching DNS (and while we're at it, probably also IP) identifiers that differ up to case.

felixfontein:
Out of curiosity, which CA(s) do(es) exhibit this normalization behavior?

lspiehler:
I'm observing the behavior with Let's Encrypt. It's the only CA I've tested so far. What do you think about also forcing the "authorizations" keys to be lowercase?

lspiehler:
I ran into the issue you described while testing the ACME server I developed for https://PKIaaS.io. I noticed there's another place the authorizations dict is defined for ACMEv2 in the load_authorizations function. After making the keys lowercase there as well, I can confirm this change is working successfully for Let's Encrypt, ZeroSSL, BuyPass, and PKIaaS.io. Let me know if there's anything else you'd like me to test. Thanks!

lspiehler:
@felixfontein, have you had a chance to review the latest commit? I think I've addressed the concerns you've mentioned. Let me know if there's anything additional I can do. Thanks!

felixfontein:
@lspiehler the PR is still on my review list, but currently I don't have much time. I should be able to take a closer look later this month, probably in the next week or the week thereafter.

plugins/modules/acme_certificate.py
```diff
@@ -835,10 +835,10 @@ def get_certificate(self):
         with an error.
         '''
         for identifier_type, identifier in self.identifiers:
-            authz = self.authorizations.get(combine_identifier(identifier_type, identifier))
+            authz = self.authorizations.get(combine_identifier(identifier_type, identifier.lower()))
             if authz is None:
                 raise ModuleFailException('Found no authorization information for "{identifier}"!'.format(
-                    identifier=combine_identifier(identifier_type, identifier)))
+                    identifier=combine_identifier(identifier_type, identifier.lower())))
```
felixfontein, line 841: This line should not have a change.

felixfontein (commented):
I did find some time to review this. There are three issues:

1. Instead of calling `.lower()` all over the place, add a function `normalize_combined_identifier` which converts the identifier part to lower-case, and use that one.
2. Your patch is potentially modifying the return values of the module in some places. In line 968 of modules/acme_certificate.py, change `split_identifier(k)[1]` to `v.identifier` (you can get rid of `k` completely); the same is true for lines 758 and 759, where you can use `authz.identifier` instead of `identifier`.
3. You don't normalize in line 786 of modules/acme_certificate.py.

plugins/modules/acme_certificate.py
lspiehler: Firstly, thank you for your time and patience. I'm hoping I understood all of the issues you mentioned and your suggestions. I believe I've addressed all three. I've also tested again with the same four CAs I tested previously (Let's Encrypt, ZeroSSL, BuyPass, and PKIaaS.io) using multiple SANs with some lower case only and some mixed case. All tests were successful. Let me know if there's anything else I can do. Thanks!

plugins/module_utils/acme/challenges.py
```diff
@@ -47,6 +47,11 @@ def combine_identifier(identifier_type, identifier):
     return '{type}:{identifier}'.format(type=identifier_type, identifier=identifier)
 
 
+def normalize_combined_identifier(identifier):
+    parts = split_identifier(identifier)
+    return '{type}:{identifier}'.format(type=parts[0], identifier=parts[1].lower())
```
felixfontein, line 52: I would split this up a bit more (and also use `combine_identifier`):
```suggestion
    identifier_type, identifier = split_identifier(identifier)
    # Normalize DNS names and IPs
    identifier = identifier.lower()
    return combine_identifier(identifier_type, identifier)
```

felixfontein (commented):
Can you please also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

lspiehler:
I committed your suggestion and included a changelog fragment. Thanks for all your help!

changelogs/fragments/803-fix-authorization-failure-with-mixed-case-sans.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - acme_certificate - fix authorization failure when CSR contains SANs with mixed case
```
felixfontein, line 2:
```suggestion
  - acme_certificate - fix authorization failure when CSR contains SANs with mixed case (https://github.com/ansible-collections/community.crypto/pull/803).
```

felixfontein (commented):
One last small change :)

lspiehler:
Suggestion committed!

felixfontein (approved):

felixfontein:
@lspiehler thanks a lot for reporting and fixing this!

felixfontein:
@lspiehler community.crypto 2.22.2 is out including this fix.

PR #844, opened 2025-01-29
luks_device: allow passphrases to contain newlines

ilia-kats:
#### SUMMARY
When passing a passphrase through stdin, add `--key-file=-` to the command-line arguments (and never append a newline to the data passed to the command). This is useful when passing binary keyfiles from an ansible vault, as it removes the restriction that the binary data cannot contain newlines.

The only exception is adding a new key to an existing container, as in that case the two passphrases are separated by a new line.

I've only tested opening a container with a binary keyfile containing newlines (given as base64-encoded string) in a VM, but the unit tests in `luks_device.py` all pass.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

luks_device


felixfontein:
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) and an integration test that covers all operations that should work?

Also in `man cryptsetup` there's the following note about reading passphrases:

> From stdin: LUKS will read passphrases from stdin up to the first newline character or the compiled-in maximum key file length. If --keyfile-size is given, it is ignored.

That doesn't sound like the PR will actually work.

ilia-kats:
> On the other hand, it says about reading passphrases from a keyfile:
> 
> > From a key file: It will be truncated to the key size of the used cipher or the size given by -s and directly used as a binary key.
> > WARNING: The --hash argument is being ignored. The --hash option is usable only for stdin input in plain mode.
> > If the key file is shorter than the key, cryptsetup will quit with an error. The maximum input size is defined by the same compiled-in default as for the maximum key file size and can be overwritten using --keyfile-size option.
> 
> This sounds like the PR might break all existing users that don't use keyfiles, since what's passed in via `--key-file=-` is treated differently from what is passed in as a regular passphrase.

If I'm reading the manpage correctly, that only applies to plain mode, which is never used by the `luks_device` module, only the LUKS mode is used. As mentioned, I successfully tested this in a VM that had an encrypted device created manually with a keyfile on disk and could be succesfully opened by Ansible with this change.

I've added an integration test and a changelog fragment.

felixfontein:
Yep, it definitely does not apply to LUKS, but to plain. I was apparently not awake enough this morning...

Thanks for adding the test, and figuring out how to make changing the key work also for passphrases with newlines in them :)

Regarding the checked in key `keyfile3`: how about replacing it by something reproducible, for example `dd if=/dev/zero of=/dev/stdout bs=1 count=4096 | openssl enc -aes-256-ctr -pass pass:1234 -nosalt`? (And mention the command used to create it, that can be used to re-create it, in say `keyfile3.txt`, or a `README` in that directory?) I don't like the idea of having random binary blobs in the repository :)

changelogs/fragments/luks_device_passphrase_newlines.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - luks_device - allow passphrases to contain newlines (https://github.com/ansible-collections/community.crypto/pull/844)
```
felixfontein, line 2:
```suggestion
  - luks_device - allow passphrases to contain newlines (https://github.com/ansible-collections/community.crypto/pull/844).
```

ilia-kats:
Done, thanks for the suggestion. Note that the CI forced me to add a licence header to `keyfile3.txt`, not sure if this is ghe way to go or if there is a way around that.

felixfontein:
> Done, thanks for the suggestion. Note that the CI forced me to add a licence header to `keyfile3.txt`, not sure if this is ghe way to go or if there is a way around that.

That's the way to go. Every file in the repo should have a license assigned so that it is reuse compliant. (Right now that's not the case for a few `.pem` files, I have no idea what to do there...)

plugins/modules/luks_device.py
```diff
@@ -706,14 +705,16 @@ def run_luks_add_key(self, device, keyfile, passphrase, new_keyfile,
         if keyfile:
             args.extend(['--key-file', keyfile])
         else:
+            args.extend(['--key-file', '-', '--keyfile-size', str(len(passphrase))])
```
felixfontein, line 708: That's a nice trick to make this work!

felixfontein (approved):
LGTM!

felixfontein:
@ilia-kats thanks for improving the module!

PR #848, opened 2025-02-17
openssl_pkcs12: Add support for `certificate_content` and `other_certificates_content`

### LINKED ISSUE #847: Support cert(s) from content in the pkcs12 module
The module supports `privatekey_content` but not `certificate_content` -- would we be open to adding such an option, then I can see that we get this added.

apollo13:
Fixes #847

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
openssl_pkcs21

### DIFF
--- a/changelogs/fragments/openssl_pkcs12_content.yml
+++ b/changelogs/fragments/openssl_pkcs12_content.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - openssl_pkcs12 - the module now supports ``certificate_content``/``other_certificates_content`` for cases where the data already exists in memory and not yet in a file (https://github.com/ansible-collections/community.crypto/issues/847, https://github.com/ansible-collections/community.crypto/pull/848).
--- a/plugins/modules/openssl_pkcs12.py
+++ b/plugins/modules/openssl_pkcs12.py
@@ -55,15 +55,15 @@ options:
   other_certificates_content:
     description:
       - List of other certificates to include.
-      - Assumes there is one PEM-encoded certificate per item.
+      - Assumes there is one PEM-encoded certificate per item. If an item contains multiple PEM certificates, set O(other_certificates_parse_all)
       - Mutually exclusive with O(other_certificates).
     type: list
     elements: str
     version_added: "2.26.0"
   other_certificates_parse_all:
     description:
-      - If set to V(true), assumes that the files mentioned in O(other_certificates) can contain more than one certificate
-        per file (or even none per file).
+      - If set to V(true), assumes that the files mentioned in O(other_certificates)/O(other_certificates_content) can contain more than one
+        certificate per file/item (or even none per file/item).
     type: bool
     default: false
     version_added: 1.4.0
@@ -280,6 +280,7 @@ pkcs12:
 
 import abc
 import base64
+import itertools
 import os
 import stat
 import traceback
@@ -430,8 +431,11 @@ class Pkcs(OpenSSLObject):
                     load_certificate(other_cert, backend=self.backend) for other_cert in self.other_certificates
                 ]
         elif self.other_certificates_content:
+            certs = self.other_certificates_content
+            if self.other_certificates_parse_all:
+                certs = list(itertools.chain.from_iterable(split_pem_list(content) for content in certs))
             self.other_certificates = [
-                load_certificate(None, content=to_bytes(other_cert), backend=self.backend) for other_cert in self.other_certificates_content
+                load_certificate(None, content=to_bytes(other_cert), backend=self.backend) for other_cert in certs
             ]
 
     @abc.abstractmethod
--- a/plugins/modules/openssl_pkcs12.py
+++ b/plugins/modules/openssl_pkcs12.py
@@ -48,9 +48,18 @@ options:
       - List of other certificates to include. Pre Ansible 2.8 this parameter was called O(ca_certificates).
       - Assumes there is one PEM-encoded certificate per file. If a file contains multiple PEM certificates, set O(other_certificates_parse_all)
         to V(true).
+      - Mutually exclusive with O(other_certificates_content).
     type: list
     elements: path
     aliases: [ca_certificates]
+  other_certificates_content:
+    description:
+      - List of other certificates to include.
+      - Assumes there is one PEM-encoded certificate per item.
+      - Mutually exclusive with O(other_certificates).
+    type: list
+    elements: str
+    version_added: "2.26.0"
   other_certificates_parse_all:
     description:
       - If set to V(true), assumes that the files mentioned in O(other_certificates) can contain more than one certificate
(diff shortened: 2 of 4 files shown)


plugins/modules/openssl_pkcs12.py
```diff
@@ -402,6 +426,10 @@ def __init__(self, module, backend, iter_size_default=2048):
                 self.other_certificates = [
                     load_certificate(other_cert, backend=self.backend) for other_cert in self.other_certificates
                 ]
+        elif self.other_certificates_content:
+            self.other_certificates = [
+                load_certificate(None, content=to_bytes(other_cert), backend=self.backend) for other_cert in self.other_certificates_content
```
apollo13, line 431: I wonder if `load_certificate` should do a `to_bytes` instead.

tests/integration/targets/openssl_pkcs12/tasks/impl.yml
```diff
@@ -4,6 +4,51 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 - block:
+  - name: "({{ select_crypto_backend }}) Generate PKCS#12 file from contents"
```
apollo13, line 7: I just added this minimal test because the other tests basically test the new behavior as well (since everything is basically redirect to `content` now). What worries me a bit about the tests is that I don't understand how for instance 
```
name: "({{ select_crypto_backend }}) Generate PKCS#12 file with multiple certs and passphrase, again (idempotency)"
```
is actually checking idempotency. It does register `p12_multiple_certs_idempotency` but never checks that the result is unchanged (cannot find that the variable is used at all)?

changelogs/fragments/openssl_pkcs12_content.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - openssl_pkcs12 - The module now supports ``certificate_content``/``other_certificates_content`` to match ``privatekey_content`` (https://github.com/ansible-collections/community.crypto/pull/848).
```
felixfontein, line 2: I would replace "to match `privatekey_content`" with something that explains what you use them for. `privatekey_content` is there to avoid having to write private keys to disk. You don't need to avoid writing certificates to disk for security reasons.
```suggestion
  - openssl_pkcs12 - the module now supports ``certificate_content``/``other_certificates_content`` to match ``privatekey_content`` (https://github.com/ansible-collections/community.crypto/issues/847, https://github.com/ansible-collections/community.crypto/pull/848).
```

plugins/modules/openssl_pkcs12.py
```diff
@@ -51,6 +51,13 @@
     type: list
     elements: path
     aliases: [ca_certificates]
+  other_certificates_content:
+    description:
+      - List of other certificates to include.
+      - Mutually exclusive with O(other_certificates).
```
felixfontein, line 57: The documentation of `other_certificates` should say that it's mutually exclusive with this option.

plugins/modules/openssl_pkcs12.py
```diff
@@ -63,6 +70,12 @@
       - The path to read certificates and private keys from.
       - Must be in PEM format.
     type: path
+  certificate_content:
+    description:
+      - Content of the certificate file.
+      - Mutually exclusive with O(certificate_path).
```
felixfontein, line 76: Same here.

plugins/modules/openssl_pkcs12.py
```diff
@@ -51,6 +51,13 @@
     type: list
     elements: path
     aliases: [ca_certificates]
+  other_certificates_content:
+    description:
+      - List of other certificates to include.
```
felixfontein, line 56: I think this needs similar explanations as `other_certificates` for what the content is and about `other_certificates_parse_all`.

plugins/modules/openssl_pkcs12.py
```diff
@@ -63,6 +70,12 @@
       - The path to read certificates and private keys from.
       - Must be in PEM format.
     type: path
+  certificate_content:
+    description:
+      - Content of the certificate file.
```
felixfontein, line 75:
```suggestion
      - Content of the certificate file in PEM format.
```

plugins/modules/openssl_pkcs12.py
```diff
@@ -402,6 +426,10 @@ def __init__(self, module, backend, iter_size_default=2048):
                 self.other_certificates = [
                     load_certificate(other_cert, backend=self.backend) for other_cert in self.other_certificates
                 ]
+        elif self.other_certificates_content:
+            self.other_certificates = [
+                load_certificate(None, content=to_bytes(other_cert), backend=self.backend) for other_cert in self.other_certificates_content
+            ]
```
felixfontein, line 432: This is missing support for `other_certificates_parse_all`.

tests/integration/targets/openssl_pkcs12/tasks/impl.yml
felixfontein: Did you check `tests/integration/targets/openssl_pkcs12/tests/validate.yml`?

apollo13: Outch, that is embarrassing, I saw some asserts in `impl.yml` like:
```
  - name: "({{ select_crypto_backend }}) Validate PKCS#12"
    assert:
      that:
      - p12_standard.pkcs12 == ansible_p12_content.content
      - p12_standard_idempotency.pkcs12 == p12_standard.pkcs12
```

and assumed the tests where inline. Sorry for that!

plugins/modules/openssl_pkcs12.py
apollo13: I thought about this when writing the code, does it actually make sense to have for the content variant? Wouldn't be hard to add but probably not common?

apollo13:
Will add some further tests later on.

tests/integration/targets/openssl_pkcs12/tasks/impl.yml
felixfontein: Using `tests/validate.yml` for assertions is something that community.crypto (resp. its tests) have been doing for a loong time. I can't say I like it too much, but also not that little that I'd change all of them ;-) Some newer modules simply have all tests in `tasks/main.yml` (or some files in `tasks/`)...

plugins/modules/openssl_pkcs12.py
felixfontein: No idea... One possible use case would be loading the CA file (containing a list of PEMs) on the controller and running the module on the remote. If you can use that option here as well, you don't need to write the file to disk on the remote first.

apollo13: Added, since it doesn't take to much effort. Let's see what the tests say.

apollo13:
@felixfontein Looks good now I think, ready to ship from my end.

plugins/modules/openssl_pkcs12.py
```diff
@@ -342,8 +359,9 @@ def load_certificate_set(filename, backend):
     '''
     Load list of concatenated PEM files, and return a list of parsed certificates.
     '''
-    with open(filename, 'rb') as f:
-        data = f.read().decode('utf-8')
+    if filename:
+        with open(filename, 'rb') as f:
+            data = f.read().decode('utf-8')
     return [load_certificate(None, content=cert.encode('utf-8'), backend=backend) for cert in split_pem_list(data)]
```
felixfontein, line 365: Where does `data` come from if `filename` is falsy?

(I'm surprised the linters aren't reporting anything.)

felixfontein:
@apollo13 thanks for your contribution!

PR #862, opened 2025-04-21
Replace vendored certificates with self-created certificates of similar structure

felixfontein:
##### SUMMARY
- [x] Manage to re-create the certs.
- [x] Anonymize certs so that original properties will not be recognizable anymore.
- [x] Maybe try to make the whole process reproducable. For that we likely have to vendor the keys (in some form that stupid checkers don't start screaming :roll_eyes:), or find another way to make key generation reproducible. Some signatures might also be randomized.
- [ ] ~~Add `.license` files, and take over other changes from #509.~~ Let's do that in the other PR.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
integration tests


felixfontein:
(Regenerating the certificates now produces slightly different files for some of them, but that's unavoidable since signing sometimes uses randomness - depending on the algorithm used. In any case, everything that's verified in the tests works fine even after regeneration.)

PR #904, opened 2025-05-29
Document supported curves for Elliptic Curve keys on ACME Accounts

ziegenberg:
##### SUMMARY

It took me way too long to figure out why I was greeted with `Error while parsing account key: unknown elliptic curve: brainpoolP256r1`, so I'm now documenting this properly for everybody else.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

acme_account

##### ADDITIONAL INFORMATION

none

### DIFF
--- a/plugins/doc_fragments/_acme.py
+++ b/plugins/doc_fragments/_acme.py
@@ -82,6 +82,7 @@ options:
   account_key_src:
     description:
       - Path to a file containing the ACME account RSA or Elliptic Curve key.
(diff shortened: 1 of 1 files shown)


felixfontein:
The unrelated sanity tests are now fixed (#905); the other failing sanity tests are caused by the missing quotes (you are unintentionally adding a YAML dictionary with a single long key and a value to the list of paragraphs :) ).

PR #928, opened 2025-07-04
Ensure consistent SSH key format with idempotent Ed25519 key regeneration

theputta:
##### SUMMARY
- Added task to generate Ed25519 SSH keypair at /home/devops/.ssh/id_ed25519
- Enforces OpenSSH private key format using `private_key_format: ssh`
- Uses `regenerate: full_idempotence` to only replace key when format or options mismatch

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
  community.crypto.openssh_keypair:

### DIFF
--- a/plugins/modules/openssh_keypair.py
+++ b/plugins/modules/openssh_keypair.py
@@ -158,7 +158,6 @@ EXAMPLES = r"""
   community.crypto.openssh_keypair:
     path: /tmp/id_ssh_rsa
     force: true
-    
 - name: Regenerate SSH keypair only if format or options mismatch
   community.crypto.openssh_keypair:
     path: /home/devops/.ssh/id_ed25519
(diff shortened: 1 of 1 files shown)


plugins/modules/openssh_keypair.py
```diff
@@ -158,6 +158,13 @@
   community.crypto.openssh_keypair:
     path: /tmp/id_ssh_rsa
     force: true
+
```
felixfontein, line 161: This line contains trailing whitespace:
```suggestion

```

felixfontein:
Hmm, I think my suggestion accidentally removed the whole line, instead of simply removing the trailing whitespace. Can you re-add a line-break (but without trailing whitespace)? Thanks.

felixfontein:
Closing in favor of #932.

PR #951, opened 2025-08-10
[WIP] Use return doc fragments

felixfontein:
##### SUMMARY
Use https://github.com/ansible/ansible/pull/72635.

This shouldn't be merged before ansible-core 2.20 is out, or maybe even only later, since it will break documentation for ansible-core verisons before 2.20:

ansible-core 2.17:
```
ERROR! Unable to retrieve documentation from 'community.crypto.openssl_csr_pipe' due to: dictionary update sequence element #0 has length 28; 2 is required. dictionary update sequence element #0 has length 28; 2 is required
```

ansible-core 2.18:
```
ERROR! Unable to retrieve documentation from 'community.crypto.openssl_csr_pipe'. dictionary update sequence element #0 has length 28; 2 is required
```

ansible-core 2.19:
```
[ERROR]: Unable to retrieve documentation from 'community.crypto.openssl_csr_pipe': dictionary update sequence element #0 has length 28; 2 is required
```

@bcoca how about changing older ansible-core branches so that `extends_documentation_fragment` (with a string or list value) doesn't cause a total breakdown, but is simply ignored, or (even better) causes a warning that return documentation is incomplete and requires a newer ansible-core version to render completely?

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
openssl_privatekey* modules
openssl_csr* modules

### DIFF
--- a/antsibull-nox.toml
+++ b/antsibull-nox.toml
@@ -57,6 +57,7 @@ codeblocks_restrict_types = [
 codeblocks_restrict_type_exact_case = true
 codeblocks_allow_without_type = false
 codeblocks_allow_literal_blocks = false
+ansible_core_package = "https://github.com/ansible/ansible/archive/devel.tar.gz"
 
 [sessions.license_check]
 run_reuse = true
--- a/plugins/doc_fragments/_module_csr.py
+++ b/plugins/doc_fragments/_module_csr.py
@@ -398,3 +398,11 @@ name_constraints_excluded:
   elements: str
   sample: ['email:.com']
 """
+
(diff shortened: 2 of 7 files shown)


felixfontein:
I created https://github.com/ansible/ansible/pull/85638 to fix the failing sanity tests in ansible-core devel.

bcoca:
By default ansible-doc should ignore 'unknown' entries. It was originally designed to allow users to create their own internal conventions for their own internal contexts.  This should be a 'passthrough' and not an error, open an issue and I'll address it, but I cannot guarantee how far it will be backported if accepted at all.

PR #966, opened 2025-11-04
Support custom X509 extensions when generating a CSR

### LINKED ISSUE #401: openssl_csr: allow to add arbitrary extensions
##### SUMMARY
See https://github.com/ansible/ansible/issues/61473 for details.

The idea is to allow to specify arbitrary extension OIDs with values and critical flags. The module should add them to the CSR (resp. ensure they are present in the CSR), and make sure they don't collide with extensions added by the module itself.

Now that cryptography allows to serialize and deserialize arbitrary extensions, we can properly implement it.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
openssl_csr

zajdee:
##### SUMMARY
This change implements the support of custom X509 extensions when generating a CSR.

Due to the need for an ASN.1 encoder, this change pulls a new component, `pyasn1` ([Github page](https://pypi.org/project/pyasn1/)), into the project. Encoder compatibility with `pyasn1` versions back to `0.4.8` was tested. This extra Python component is only needed when `custom_extensions` are specified.

Because there can be various scenarios when generating the certificate request using Ansible, the values can be provided as raw data, raw Base64-encoded data, or as one of the following supported data types: `int`, `real` (`float`), `bool`, and `str`.

Empty extension value is not supported, and if specified, it either breaks the run, or (if allowed) such an extension definition is skipped.

Fixes #401

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
The CSR generating backend, docs, and tests were modified.
 
##### ADDITIONAL INFORMATION

A potential use of this new feature is as follows:

```yaml
- name: Generate mTLS CSR
  community.crypto.openssl_csr:
(...)
  custom_extensions:
      - oid: 1.3.6.1.4.1.34380.1.1.12
        value: "DE"
        critical: false
        skip_if_empty: true
      - oid: 1.3.6.1.4.1.34380.1.1.13
        value: ""
        critical: false
        skip_if_empty: true
      - oid: 1.3.6.1.4.1.34380.1.1.14
        value: 17
        value_type: int
        critical: false
      - oid: 1.3.6.1.4.1.34380.1.1.15
        value: 3.14
        value_type: real
        critical: false
      - oid: 1.3.6.1.4.1.34380.1.1.16
        value: true
        value_type: bool
        critical: false
      - oid: 1.3.6.1.4.1.34380.1.1.17
        value: false
        value_type: bool
        critical: false
      - oid: 1.3.6.1.4.1.34380.1.1.18
        value_raw: "DE"
        critical: false
        skip_if_empty: true
      - oid: 1.3.6.1.4.1.34380.1.1.19
        value_b64: "REU=" # DE
        critical: true
        skip_if_empty: true
```


plugins/doc_fragments/_module_csr.py
```diff
@@ -325,6 +326,53 @@ class ModuleDocFragment:
...
+        description:
+          - The type of the value. Only valid if O(custom_extensions[].value) is specified.
+          - 'Supported data types are: V(str), V(bool), V(int), V(real).'
+        type: str
+        default: str
+      value:
+        description:
+          - The value of the custom extension.
+          - Mutually exclusive with O(custom_extensions[].value_raw) and O(custom_extensions[].value_b64).
+        type: str
```
felixfontein: I'd use
```suggestion
        type: raw
```
here. That way inputs are not forcible converted to strings before they can be handled.

plugins/doc_fragments/_module_csr.py
```diff
@@ -325,6 +326,53 @@ class ModuleDocFragment:
...
+      skip_if_empty:
+        description:
+          - Allow empty value to be specified. In such case the extension will be skipped when processing.
+        type: bool
+        default: false
+      value_type:
+        description:
+          - The type of the value. Only valid if O(custom_extensions[].value) is specified.
+          - 'Supported data types are: V(str), V(bool), V(int), V(real).'
```
felixfontein: I would remove `V(real)`. Handling strings, booleans, and integers should be OK.

plugins/doc_fragments/_module_csr.py
```diff
@@ -325,6 +326,53 @@ class ModuleDocFragment:
...
+        default: str
+      value:
+        description:
+          - The value of the custom extension.
+          - Mutually exclusive with O(custom_extensions[].value_raw) and O(custom_extensions[].value_b64).
+        type: str
+      value_raw:
+        description:
+          - The raw value of the custom extension. Will be copied to the extension as-is.
+          - The value is always assumed to be an UTF8String
```
felixfontein: What's the difference of using `value` with `value_type=str`?

plugins/module_utils/_crypto/module_backends/csr.py
```diff
@@ -185,6 +264,10 @@ def __init__(self, *, module: AnsibleModule) -> None:
         self.crl_distribution_points: (
             list[cryptography.x509.DistributionPoint] | None
         ) = None
+        self.custom_extensions: (
+            list[tuple[cryptography.x509.UnrecognizedExtension, bool]] | None
+        ) = None
```
felixfontein: I would remove the `| None` and set it to `[]`. There's no need to distinguish between `None` and `[]`.

plugins/module_utils/_crypto/module_backends/csr.py
```diff
@@ -906,6 +1001,21 @@ def get_csr_argument_spec() -> ArgumentSpec:
                 "mutually_exclusive": [("full_name", "relative_name")],
                 "required_one_of": [("full_name", "relative_name", "crl_issuer")],
             },
+            "custom_extensions": {
+                "type": "list",
+                "elements": "dict",
+                "options": {
+                    "critical": {"type": "bool", "default": False},
+                    "oid": {"type": "str"},
```
felixfontein: Shouldn't the `oid` be required?
```suggestion
                    "oid": {"type": "str", "required": True},
```

felixfontein (commented):
Thanks again for your contribution! I've had time for a first review pass.

felixfontein:
Ping @zajdee

PR #1057, opened 2026-07-12
Update vendored OID names list

felixfontein:
##### SUMMARY
Update to the latest version of https://github.com/openssl/openssl/blob/master/crypto/objects/objects.txt.

Note that we made sure that old names supported before are still supported. When OIDs are converted to names, some strings are now changing though:
1. OIDs that were unknown before and now have a name change from a dotted number to a name.
2. OIDs that have new names now use the new names. (The old names are still supported when parsing though.)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
OID names


felixfontein:
The CI failure in the extra sanity tests is due to https://github.com/ansible-community/antsibull-nox/issues/213.

